PR #4, opened 2020-09-17
IA-2209 Add deleting staging buckets in resource validator

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2209

Note conclusion around https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/807436289/2020-09-17+Leonardo+Async+Processes will likely change the PR...But regardless the conclusion, we should be able to move `BucketRemover` to proper place relatively easily

* Some refactoring to make adding `BucketRemover` easier
* Address comments from last PR review
* Fix some package naming introduced by accidental IJ refactoring
* `BucketRemover` is the real addition
* Added `KubernetesClusterRemover`. It doesn't really remove the cluster, but it'll detect clusters that need to be removed, and alert us in team channel when it happens. This should be good intermediate step before we come to a conclusion around [this](https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/807436289/2020-09-17+Leonardo+Async+Processes)

`BucketRemover` is validated in dev. Sample log:
```
{"@timestamp":"2020-09-17T20:54:58.526Z","@version":"1","message":"Successfully deleted all objects in leostaging-saturn-19f4e265-3b350c0d-58cb-468a-b7df-5446f2b46d49","logger_name":"com.broadinstitute.dsp.resourceValidator.ResourceValidator","thread_name":"scala-execution-context-global-247","severity":"INFO","level_value":20000}
{"@timestamp":"2020-09-17T20:54:58.585Z","@version":"1","message":"Successfully deleted all objects in leostaging-saturn-1f4cfa96-1d1f26da-7bdd-438a-b922-f9039f2c0d99","logger_name":"com.broadinstitute.dsp.resourceValidator.ResourceValidator","thread_name":"scala-execution-context-global-251","severity":"INFO","level_value":20000}
...
{"@timestamp":"2020-09-17T21:08:34.787Z","@version":"1","message":"remove-staging-buckets | Check out gs://qi-test/remove-staging-buckets/dry-run-2020-09-17T20:44:00.557Z for more details","logger_name":"com.broadinstitute.dsp.resourceValidator.ResourceValidator","thread_name":"scala-execution-context-global-252","severity":"WARN","level_value":30000}
```

13351 buckets are removed from dev run


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -38,5 +41,30 @@ object DbReader {
...
+
+    // Return all clusters that has no non "deleted" nodepool
+    override def getK8sClustersToDelete: Stream[F, KubernetesClusterId] =
+      sql"""SELECT kc.id, kc.location, kc.clusterName
+            FROM KUBERNETES_CLUSTER kc 
+            WHERE NOT EXISTS (
+              SELECT *
+                FROM KUBERNETES_CLUSTER kc 
+                LEFT JOIN NODEPOOL as np on np.clusterId=kc.id
+              WHERE np.status != "DELETED"
```
kyuksel, line 59: May need to include `ERROR` as well to catch cases where nodepool deletion errored out or some other unexpected issues came up.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -38,5 +41,30 @@ object DbReader {
...
+        .query[BucketToRemove]
+        .stream
+        .transact(xa)
+
+    // Return all clusters that has no non "deleted" nodepool
+    override def getK8sClustersToDelete: Stream[F, KubernetesClusterId] =
+      sql"""SELECT kc.id, kc.location, kc.clusterName
+            FROM KUBERNETES_CLUSTER kc 
+            WHERE NOT EXISTS (
+              SELECT *
```
kyuksel, line 56: I wonder if `SELECT *` causes a lot of data to be read out of the DB or if the DB engine is smart enough to optimize it out. If the former, we may want to specify a column instead of `*` since we don't care about the content.

kyuksel: It looks like it's [the latter](https://dev.mysql.com/doc/refman/8.0/en/exists-and-not-exists-subqueries.html)

>Traditionally, an EXISTS subquery starts with SELECT *, but it could begin with SELECT 5 or SELECT column1 or anything at all. MySQL ignores the SELECT list in such a subquery, so it makes no difference.

so you can ignore my previous comment 😊

Qi77Qi: I'm not a SQL expert..this is good to know :)

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
Qi77Qi: it looks like nodepool can end up in `ERROR` state, so I don't think it should be here. I'll make another checker similar to https://github.com/broadinstitute/leonardo-cron-jobs/blob/master/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeChecker.scala

kyuksel: This feels a bit different than what `ErroredRuntimeChecker` does because we're interested not only in cleaning up errored nodepools but also the cluster they belong to, if there are no non-deleted and non-errored nodepools on that cluster. Also happy to see how you want to implement it differently, though.

Qi77Qi: It's similar to `ErroredRuntimeChecker` only in a sense that one cares about errored runtime, and one cares about errored nodepool...The action taken will be different in many regards as well, the only similar part is if we detect `ERROR` in google for nodepool, we don't want to delete k8s cluster (similar to we don't want to delete errored runtime)

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedDiskChecker.scala
```diff
@@ -0,0 +1,37 @@
...
+// Implements CheckRunner[F[_], A]
+object DeletedDiskChecker {
+  def impl[F[_]: Timer](
+    dbReader: DbReader[F],
+    deps: DeletedDiskCheckerDeps[F]
+  )(implicit F: Concurrent[F], logger: Logger[F], ev: ApplicativeAsk[F, TraceId]): CheckRunner[F, Disk] =
+    new CheckRunner[F, Disk] {
+      override def configs = CheckRunnerConfigs("deleted-disks", true)
+      override def dependencies: CheckRunnerDeps[F] = deps.checkRunnerDeps
+      override def aToScan: fs2.Stream[F, Disk] = dbReader.getDeletedDisks
```
kyuksel, line 20: I'm guessing `a` refers to a type here. I wonder if it'd be more understandable to rename the corresponding methods as `resourcesToScan` and `checkResource`.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -56,12 +57,16 @@ object DbReader {
               SELECT *
                 FROM KUBERNETES_CLUSTER kc 
                 LEFT JOIN NODEPOOL as np on np.clusterId=kc.id
-              WHERE np.status != "DELETED"
+              WHERE np.status != "DELETED" and np.status !="ERROR"
```
kyuksel, line 60: 👍  

Can we update the comment

>// Return all clusters that has no non "deleted" nodepool

to reflect this change as well?

kyuksel (approved):
Looks good 👍  A couple of comments with method renaming suggestions and a comment fix are inline.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -38,5 +42,34 @@ object DbReader {
...
+
+    override def getBucketsToDelete: Stream[F, BucketToRemove] =
+      sql"""select googleProject, stagingBucket from CLUSTER WHERE status="Deleted" and destroyedDate < now() - interval 15 DAY;"""
+        .query[BucketToRemove]
+        .stream
+        .transact(xa)
+
+    // Return all clusters that has no non "DELETED" | `ERROR` nodepool
+    override def getK8sClustersToDelete: Stream[F, KubernetesClusterId] =
+      sql"""SELECT kc.id, kc.location, kc.clusterName
```
rtitle, line 54: We might need to iterate on this query.. I had wanted to add a configurable "grace period" during which we leave empty clusters running.

Also I think all clusters will have at least 1 nodepool at all times, so this query may never trigger any clusters.

Qi77Qi: so should the criteria be more than one nodepool?

rtitle: maybe something like:
```
select kc.id, kc.location, kc.clusterName
from KUBERNETES_CLUSTER kc
where (
  select count(*) from NODEPOOL np WHERE np.clusterId = kc.id AND np.status <> 'Error' AND np.status <> 'Deleted'
) = 1
AND kc.status <> 'Error' AND kc.status <> 'Deleted'
```

Just thinking out loud, maybe there is a better way to distinguish the dummy nodepool than just count.

kyuksel: I believe that the `NODEPOOL` table has an `isDefault` field, which may help.

Qi77Qi: @rtitle what is `kc.status <> 'Error' AND kc.status <> 'Deleted'` for? I thought we're deleting the cluster regardless of its current status as long as it's not being actively used anymore

PR #9, opened 2020-09-23
[no ticket] make staging bucket report a bit more accurate

Qi77Qi:
Minor fix:
Currently, we're blindly reporting all staging bucket candidates that are need to be deleted because to be more accurate, previously I thought we need to make an additional `getBucket` call, but now that I look at `deleteBucket` API, it's returning `true` when the deletion actually happened, and `false` if not found, hence we can tell if a bucket was deleted fairly easily.

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/BucketRemover.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/BucketRemover.scala
@@ -32,7 +32,17 @@ object BucketRemover {
         implicit ev: ApplicativeAsk[F, TraceId]
       ): F[Option[BucketToRemove]] =
         a.bucket
-          .traverse(b => deps.storageService.deleteBucket(a.googleProject, b, true).compile.drain.as(a))
+          .flatTraverse { b =>
+            deps.storageService.deleteBucket(a.googleProject, b, true).compile.last.map { resultOpt =>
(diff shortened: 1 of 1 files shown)


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/BucketRemover.scala
```diff
@@ -32,7 +32,17 @@ object BucketRemover {
         implicit ev: ApplicativeAsk[F, TraceId]
       ): F[Option[BucketToRemove]] =
         a.bucket
-          .traverse(b => deps.storageService.deleteBucket(a.googleProject, b, true).compile.drain.as(a))
+          .flatTraverse { b =>
+            deps.storageService.deleteBucket(a.googleProject, b, true).compile.last.map { resultOpt =>
```
gpcarr, line 36: how come we don't need `.drain` anymore?

gpcarr (approved):

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/BucketRemover.scala
Qi77Qi: `.drain` is saying "we don't care about the return type of `deleteBucket`, but here we do. Depending on whether it's `true` or `false`, we're doing things a bit differently

PR #11, opened 2020-10-12
IA-2245 Add DeletedNodepoolChecker

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2245

Scans through all active nodepools, and update DB accordingly when necessary
1. if nodepool doesn't exist in google anymore, mark it as DELETED and report this nodepool
2. If nodepool exist in google in ERROR, mark it as ERROR and report this nodepool
3. if nodepool exist in non ERROR state, do nothing and don't report the nodepool


zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedNodepoolChecker.scala
```diff
@@ -0,0 +1,50 @@
...
+import fs2.Stream
+import io.chrisdavenport.log4cats.Logger
+import org.broadinstitute.dsde.workbench.model.TraceId
+
+/**
+ * Scans through all active nodepools and check if they still exist in Google.
+ * If they're deleted from Google, we're updating them to `DELETED` in leonardo DB.
+ * We're also updating any kubernetes apps running on these nodepools as `DELETED`.
+ */
+object DeletedNodepoolChecker {
```
kyuksel, line 16: Should the name reflect the fact that errored nodepools are being checked as well?

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedNodepoolChecker.scala
```diff
@@ -0,0 +1,50 @@
...
+          nodepoolToReport <- nodepoolOpt match {
+            case None =>
+              (if (isDryRun) F.unit
+               else
+                 dbReader.markNodepoolAndAppStatusDeleted(nodepoolToScan.id)).as(Some(nodepoolToScan))
+            case Some(nodepool) =>
+              if (nodepool.getStatus == com.google.container.v1.NodePool.Status.ERROR) {
+                (if (isDryRun) F.unit
+                 else
+                   dbReader.markNodepoolError(nodepoolToScan.id)).as(Some(nodepoolToScan))
```
kyuksel, line 44: Shouldn't we mark apps as errored the way we do with deleted? If so, I wonder if it makes sense to have a `markNodepoolAndAppStatus()` method that takes an additional `status` parameter so it can be used for both cases.

zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DbReaderSpec.scala
```diff
@@ -30,6 +32,10 @@ class DbReaderSpec extends AnyFlatSpec with CronJobsTestSuite with IOChecker {
     check(DbReader.activeK8sClustersQuery)
   }
 
+  it should "builds activeNodepoolsQuery properly" in {
```
kyuksel, line 35: typo
```suggestion
  it should "build activeNodepoolsQuery properly" in {
```

kyuksel (commented):
Looks good overall. Added some comments/questions inline.

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedNodepoolChecker.scala
Qi77Qi: yep...realized that this morning :)

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
```diff
@@ -26,6 +29,13 @@ object DbReader {
     sql"""select id, googleProject, location, clusterName from KUBERNETES_CLUSTER where status != "DELETED" and status != "ERROR";
         """.query[K8sClusterToScan]
 
+  val activeNodepoolsQuery =
+    sql"""select np.id, cluster.googleProject, cluster.location, cluster.clusterName, np.nodepoolName from 
+         	NODEPOOL AS np INNER JOIN KUBERNETES_CLUSTER AS cluster 
+         	on cluster.id = np.clusterId 
+         	where np.status != "DELETED" and np.status != "ERROR"
+         	""".query[NodepoolToScan]
```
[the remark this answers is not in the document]
Qi77Qi: so we don't need custom mapping for case classes specifically, as long as there're `Get` defined for each individual fields (in [here](https://github.com/broadinstitute/leonardo-cron-jobs/blob/master/core/src/main/scala/com/broadinstitute/dsp/DbReaderImplicits.scala#L9))

Qi77Qi:
@kyuksel thanks for reviewing..think I addressed all comments, let me know if you have more comments

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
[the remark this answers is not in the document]
kyuksel: Oh cool - thanks

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
```diff
@@ -73,14 +73,20 @@ object DbReader {
...
+        _ <- updateAppStatusForNodepoolId(nodepoolId, "DELETED").run
       } yield ()
       res.transact(xa)
     }
 
-    def markNodepoolError(id: Long): F[Unit] = updateNodepoolStatus(id, "ERROR").run.transact(xa).void
+    def markNodepoolAndAppError(nodepoolId: Long): F[Unit] = {
+      val res = for {
+        _ <- updateNodepoolStatus(nodepoolId, "ERROR").run
+        _ <- updateNodepoolStatus(nodepoolId, "ERROR").run
```
kyuksel, line 87: ^^ looks like copy pasta 😛

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
```diff
@@ -73,14 +73,20 @@ object DbReader {
...
-        _ <- updateNodepoolStatus(id, "DELETED").run
-        _ <- updateAppStatusForNodepoolId(id).run
+        _ <- updateNodepoolStatus(nodepoolId, "DELETED").run
+        _ <- updateAppStatusForNodepoolId(nodepoolId, "DELETED").run
       } yield ()
       res.transact(xa)
     }
 
-    def markNodepoolError(id: Long): F[Unit] = updateNodepoolStatus(id, "ERROR").run.transact(xa).void
+    def markNodepoolAndAppError(nodepoolId: Long): F[Unit] = {
```
kyuksel, line 84: Is there a reason why we couldn't have a single method 
```scala
updateNodepoolAndAppStatus(nodepoolId: Long, status: Status): F[Unit]
```
that we can use for updating the DB statuses for both `Deleted` and `Error` cases? Nit: `update` reads better to me than `mark`.

Also, any reason why not to have a `Status` ADT instead of using `String`?

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedOrErroredNodepoolChecker.scala
```diff
@@ -9,11 +9,13 @@ import io.chrisdavenport.log4cats.Logger
 import org.broadinstitute.dsde.workbench.model.TraceId
 
 /**
- * Scans through all active nodepools and check if they still exist in Google.
- * If they're deleted from Google, we're updating them to `DELETED` in leonardo DB.
- * We're also updating any kubernetes apps running on these nodepools as `DELETED`.
+ * Scans through all active nodepools, and update DB accordingly when necessary
+ *
+ * - if nodepool doesn't exist in google anymore, mark it as `DELETED`, mark associated APP as `DELETED` and report this nodepool
+ * - If nodepool exist in google in ERROR, mark it as `ERROR`, mark associated APP as `ERROR`  and report this nodepool
```
kyuksel, line 15: I wonder if we should report the app as well 🤔

kyuksel (changes requested):

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
Qi77Qi: nope..I think this was just because previous I wasn't updating them at the same time. will update

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
Qi77Qi: oh man, I was thinking the function is so simple and doesn't need a separate unit test for itself...apparently I was wrong

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedOrErroredNodepoolChecker.scala
Qi77Qi: it'll be a bit more complicated if we want to report `App` as well since for now, for a given report file, we assume they're all the same type (like runtimes, disks, nodepools). so it'll require a bit more thinking how we organize things. so I think we should punt this for now.

kyuksel: Punting for now makes sense. Should we create a ticket so we don't forget?

.github/scripts/createCommit.sh
```diff
@@ -0,0 +1,38 @@
...
+  else
+    error "version $version does not match the semver scheme 'X.Y.Z(-PRERELEASE)(+BUILD)'. See help for more information."
+  fi
+}
+
+oldVersion=$(cat charts/leonardo/Chart.yaml|grep version|cut -d' ' -f 2)
+bumpVersion $oldVersion
+sed -i "s/version: .*/version: $NEW_VERSION/g" charts/leonardo/Chart.yaml
+sed -i "s/$TARGET_TAG: .*/$TARGET_TAG: $git_hash/g" charts/leonardo/values.yaml
+git commit -am "bump $TARGET_TAG version"
```
kyuksel, line 38: Nit: add new line

.github/scripts/createCommit.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
kyuksel, line 1: Is this script under `.github` directory because that's what's required for GitHub Actions?

.github/scripts/createCommit.sh
```diff
@@ -0,0 +1,38 @@
...
+
+TARGET_TAG=$1
+
+git config user.email "broadbot@broadinstitute.org"
+git config user.name "broadbot"
+git_hash=$(git rev-parse --short "$GITHUB_SHA")
+
+# Code borrowed from https://github.com/fsaintjacques/semver-tool/blob/master/src/semver
```
kyuksel, line 9: Helpful to have the reference link above. Would be even better to comment on whether the help text on the link applies to this script and if not, clarify the distinctions here as the details of what's going on don't look straightforward. May also be good to add a comment as to how this script is used.

.github/workflows/createPullRequest.yml
```diff
@@ -3,51 +3,36 @@ name: Create terra-helm Pull Request
...
-          sed -i "s/resourceValidatorImageTag: .*/resourceValidatorImageTag: $git_hash/g" charts/leonardo/values.yaml
-          git commit -am "bump version"
+      - name: update resource-validator hash
+        if: steps.resource-validator.outputs.changed == 'true'
+        run: .github/scripts/createCommit.sh resourceValidatorImageTag
+        shell: bash
+      - name: update zombie-monitor hash
+        if: steps.zombie-monitor.outputs.changed == 'true'
+        run: .github/scripts/createCommit.sh zombieMonitorImageTag
+        shell: bash
```
kyuksel, line 35: Looks like a good walkthrough topic candidate 😉

.github/workflows/unitTests.yml
```diff
@@ -0,0 +1,25 @@
...
+        run: sbt test
+      - name: Check output
+        run: |
+          echo "Pull Request Number - ${{ steps.cpr.outputs.pull-request-number }}"
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Formatting
+        run: sbt scalafmtSbtCheck scalafmtCheck test:scalafmtCheck
```
kyuksel, line 25: Nit: add new line

zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DbReaderSpec.scala
```diff
@@ -144,4 +144,27 @@ class DbReaderSpec extends AnyFlatSpec with CronJobsTestSuite with IOChecker {
       res.unsafeRunSync()
     }
   }
+
+  it should "update nodepool and App status properly" in {
```
kyuksel, line 148: Nit
```suggestion
  it should "update nodepool and app status properly" in {
```

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
```diff
@@ -73,18 +72,10 @@ object DbReader {
     override def getk8sNodepoolsToDeleteCandidate: Stream[F, NodepoolToScan] =
       activeNodepoolsQuery.stream.transact(xa)
 
-    override def markNodepoolAndAppStatusDeleted(nodepoolId: Long): F[Unit] = {
+    override def updateNodepoolAndAppStatus(nodepoolId: Long, status: String): F[Unit] = {
```
kyuksel, line 75: Should we also use a `Status` ADT the way we do in other codebases for more type safety?

kyuksel (commented):
Thanks for addressing the previous comments. A few more inline 😊

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedOrErroredNodepoolChecker.scala
Qi77Qi: on a second thought, I'm not sure if it's that useful to report `App`, if we have a specific app that we're wondering if it being updated properly, we can always go to its nodepool, and check the report to see if it was indeed updated by `janitor` (currently resource-validator).. all reports contain the source object, which is pretty easy to understand. If in the future, we have a use case where we do care about `App`, then I think it makes more sense to have a ticket...

.github/scripts/createCommit.sh
Qi77Qi: not really...I'm just following examples on docs...don't worry about changes in `.github` yet...Things don't work properly just yet. I have a separate ticket working on this. I jsut figured I'll use this as a test. But I'm going to disable it before I merge this PR

.github/workflows/unitTests.yml
Qi77Qi: what's the particular reason for adding new line btw? I never asked before, but figured I'd ask since you seem to suggest newline all the time

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
Qi77Qi: I feel ok without an ADT for now. If we publish leo models, they won't be needed here.

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedOrErroredNodepoolChecker.scala
kyuksel: I thought it could debugging a bit faster in the case of UL questions, etc. but yea, it doesn't seem like a big deal right now since we have 1-to-1 mapping between apps and nodepools.

.github/scripts/createCommit.sh
kyuksel: I see, okay. Might be good to do it as a separate PR next time 😊

.github/workflows/unitTests.yml
kyuksel: Good question. In addition to the reasons [here](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file), I also personally find the red marking GitHub adds to diffs distracting  🤷‍♂️

kyuksel (approved):

.github/scripts/createCommit.sh
Qi77Qi: yea..sorry about this...oh actually, what I was thinking is if I fix the CI, then I don't have to manually fix the PR to `terra-helm`...that'll also prove CI works properly..so think I'm gonna try to get CI working properly before I merge this one (unless it takes too long to make CI work)

.github/workflows/createPullRequest.yml
Qi77Qi: there's an item on walkthrough page about CI for this repo...so we can talk about this then

Qi77Qi:
I realize I need to look into something for aou and won't be able to get to CI work immediately...so I'm going to merge this as is and come back to CI later

PR #14, opened 2020-10-16
[IA-2237] adding init bucket checker

jdcanas:


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -37,18 +38,18 @@ object DbReader {
...
+      sql"""select distinct googleProject, clusterName, rt.cloudService from CLUSTER AS c1
+             INNER join RUNTIME_CONFIG AS rt ON c1.`runtimeConfigId`=rt.id WHERE c1.status="Deleted"
+             and NOT EXISTS (SELECT * from CLUSTER as c2 where c2.googleProject = c1.googleProject
              and c2.clusterName=c1.clusterName and (c2.status="Stopped" or c2.status="Running"));"""
         .query[Runtime]
         .stream
         .transact(xa)
 
     override def getErroredRuntimes: Stream[F, Runtime] =
-      sql"""select distinct googleProject, clusterName, rt.cloudService from CLUSTER AS c1
```
jdcanas, line 49: my intelliJ didn't like the whitespace here apparently

project/Dependencies.scala
```diff
@@ -21,7 +21,8 @@ object Dependencies {
     "com.google.cloud" % "google-cloud-compute" % "0.118.0-alpha",
     "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2,
     "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2 % Test classifier "tests",
-    "org.scalatestplus" %% "scalacheck-1-14" % "3.2.2.0" % Test
+    "org.scalatestplus" %% "scalacheck-1-14" % "3.2.2.0" % Test,
+    "org.scalatestplus" %% "mockito-3-3" % "3.2.0.0" % Test // https://github.com/scalatest/scalatestplus-selenium
```
jdcanas, line 25: I don't think its possible to create an instance of the google `Bucket` class without interacting with an API to get/create, so I needed to mock it for testing

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/InitBucketCheckerSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    it should "return bucket if bucket still exists in Google" in {
+      forAll { (bucket: InitBucketToRemove, dryRun: Boolean) =>
+        val storageService = new BaseFakeGoogleStorage {
+          override def getBucket(googleProject: GoogleProject, bucketName: GcsBucketName, bucketGetOptions: List[Storage.BucketGetOption], traceId: Option[TraceId]): IO[Option[Bucket]] =
+            IO.pure(Some(mockBucket))
+
+          override def deleteBucket(googleProject: GoogleProject, bucketName: GcsBucketName, isRecursive: Boolean, bucketSourceOptions: List[Storage.BucketSourceOption], traceId: Option[TraceId], retryConfig: RetryConfig): Stream[IO, Boolean] =
+            if (dryRun) Stream.emit(fail("this shouldn't be called")) else Stream.empty
```
Qi77Qi, line 45: this should return `Stream.emit(true).covary[IO]` in the else branch

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/InitBucketCheckerSpec.scala
```diff
@@ -0,0 +1,63 @@
...
+        val checker = InitBucketChecker.impl(dbReader, checkerDeps)
+        val res = checker.checkResource(bucket, dryRun)
+        res.unsafeRunSync() shouldBe None
+      }
+    }
+
+    it should "return bucket if bucket still exists in Google" in {
+      forAll { (bucket: InitBucketToRemove, dryRun: Boolean) =>
+        val storageService = new BaseFakeGoogleStorage {
+          override def getBucket(googleProject: GoogleProject, bucketName: GcsBucketName, bucketGetOptions: List[Storage.BucketGetOption], traceId: Option[TraceId]): IO[Option[Bucket]] =
```
Qi77Qi, line 41: I don't think we're calling `getBucket` anywhere in actual code, no? so I think we can remove mockito as well

core/src/main/scala/com/broadinstitute/dsp/models.scala
```diff
@@ -12,4 +13,14 @@ object CloudService {
 final case class Disk(id: Long, googleProject: GoogleProject, diskName: DiskName) {
   override def toString: String = s"${id}/${googleProject.value},${diskName.value}"
 }
+
+//init buckets are different than staging buckets because we store them with gs://[GcsBucketName]/
+case class InitBucketName(value: String) extends ValueObject with Product with Serializable {
```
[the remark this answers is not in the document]
jdcanas: Thats true... I'lll move the parsing logic to the case class instead of the decoder. I looked at it initially like a sanitization issue, and we usually sanitize/check data from external sources before wrapping them in a case class, but we can do it the other way too. The gs:// string is never really used though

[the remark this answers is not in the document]
Qi77Qi: since we're treating this as a separate type, so I'm seeing `InitBucketName` as the type for buckets with `gs://` prefix...but maybe we don't need a custom type at all if you're doing decoding when we're reading from DB...where's that code btw, I didn't see it

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/InitBucketCheckerSpec.scala
jdcanas: We use it here https://github.com/broadinstitute/leonardo-cron-jobs/pull/14/files#diff-7de04ba3526138e76c3434cbbf7b17a7a85d86df137ad6a7635fb1a3fe3d4012R33 .  We don't do this for the current BucketRemover, since its not an anomaly and blindly calls delete and then reacts based on the  true/false response. In this case, we can't blindly issue the call because we want to support dryRun, which needs to know whether the bucket exists without calling delete first.

core/src/main/scala/com/broadinstitute/dsp/models.scala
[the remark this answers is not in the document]
Qi77Qi: oh nvm..it's right up there....yea, I think we do need a custom type...I kinda prefer to have `InitBucketName` having `gs://` there...I still like having the decoding to check if this is valid tho 🤔

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -58,20 +59,26 @@ object DbReader {
...
     // Checker will blindly delete all buckets returned by this function. Since we've started running the cron job daily,
     // We really only need to delete any new buckets; hence we're skipping buckets whose runtimes were deleted more than 20 days ago
-    override def getBucketsToDelete: Stream[F, BucketToRemove] =
+    override def getStagingBucketsToDelete: Stream[F, BucketToRemove] =
       sql"""select googleProject, stagingBucket from CLUSTER WHERE status="Deleted" and destroyedDate < now() - interval 15 DAY and destroyedDate > now() - interval 20 DAY;"""
         .query[BucketToRemove]
         .stream
         .transact(xa)
 
+    override def getInitBucketsToDelete: Stream[F, InitBucketToRemove] =
```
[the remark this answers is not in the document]
jdcanas: Added the unit test. Not sure what you mean by Get instance? If You mean the DBReaderImplicit `Get[InitBucketName]` exists.

core/src/main/scala/com/broadinstitute/dsp/models.scala
```diff
@@ -12,4 +13,14 @@ object CloudService {
...
+  val asGcsBucketName: GcsBucketName = GcsBucketName(value)
+}
+
+final case class InitBucketToRemove(googleProject: GoogleProject, bucket: Option[InitBucketName]) {
+  override def toString: String = s"${googleProject.value},${bucket.getOrElse("null")}"
```
Qi77Qi, line 23: didn't realize initBuckets are optional 🤔

core/src/main/scala/com/broadinstitute/dsp/DbReaderImplicits.scala
```diff
@@ -16,6 +16,7 @@ object DbReaderImplicits {
   )
 
   implicit val gcsBucketNameGet: Get[GcsBucketName] = Get[String].map(GcsBucketName)
+  implicit val initBucketNameGet: Get[InitBucketName] = Get[String].map(s => InitBucketName(s.trim.subSequence(5, s.trim.length - 1).toString))
```
Qi77Qi, line 19: maybe use `temap` here? and fail if this isn't in the format of `gs://???/`

core/src/main/scala/com/broadinstitute/dsp/models.scala
jdcanas: Yeah, I'm not sure why, but there are some NULLs in the dev DB

core/src/main/scala/com/broadinstitute/dsp/DbReaderImplicits.scala
jdcanas: Yes, it does. I added a require in the most recent round of refactors I'm about to push

core/src/main/scala/com/broadinstitute/dsp/models.scala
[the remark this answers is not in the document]
jdcanas: I compromised by making the case class enforce the prefix `gs://` and suffice `/` at construction

Qi77Qi (approved):
looks great...

Qi77Qi:
PR failed on scalafmt

PR #15, opened 2020-10-16
[IA-2253] Delete K8s clusters with no apps after grace period

kyuksel:
[IA-2033](https://broadworkbench.atlassian.net/browse/IA-2033)
[Companion Leo PR
](https://github.com/DataBiosphere/leonardo/pull/1638)

The ResourceValidator changes (soon to be its own project `janitor`™️ ) are to
- update the DB query to correctly detect K8s clusters where no apps have run for grace period (currently set to `1 hour`)
- publish a `DeleteKubernetesClusterMessage` for back Leo to process

Testing:
- Manually tested locally against dev Leo
- Unit tests being added...


project/Dependencies.scala
```diff
@@ -9,6 +9,7 @@ object Dependencies {
     "net.logstash.logback" % "logstash-logback-encoder" % "6.2",
     "ch.qos.logback" % "logback-classic" % "1.2.3",
     "ch.qos.logback" % "logback-core" % "1.2.3",
+    "com.beachape" %% "enumeratum" % "1.5.13",
```
Qi77Qi, line 12: I don't think we need this

core/src/main/scala/com/broadinstitute/dsp/models.scala
```diff
@@ -14,3 +18,29 @@ final case class Disk(id: Long, googleProject: GoogleProject, diskName: DiskName
...
+}
+
+object JsonCodec {
+  implicit val googleProjectEncoder: Encoder[GoogleProject] = Encoder.encodeString.contramap(_.value)
+}
+
+object LeoPubsubCodec {
+  import JsonCodec._
+
+  sealed trait LeoPubsubMessageType extends EnumEntry with Serializable with Product {
```
Qi77Qi, line 35: I don't think we need this

core/src/main/scala/com/broadinstitute/dsp/models.scala
```diff
@@ -14,3 +18,29 @@ final case class Disk(id: Long, googleProject: GoogleProject, diskName: DiskName
...
+    def asString: String
+    override def toString = asString
+  }
+
+  implicit val leoPubsubMessageTypeEncoder: Encoder[LeoPubsubMessageType] = Encoder.encodeString.contramap(_.asString)
```
Qi77Qi, line 40: I don't think we need this

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -64,19 +63,31 @@ object DbReader {
...
+    // or errored outside the grace period (1 hour)
+    //
+    // We are excluding clusters with only default nodepools running on them so we do not remove batch-pre-created clusters.
+    // We are calculating the grace period for cluster deletion assuming that the following are valid proxies for an app's last activity:
+    //    1. destroyedDate for deleted apps
+    //    2. dateAccessed for error'ed apps
+    override def getKubernetesClustersToDelete: Stream[F, KubernetesClusterToRemove] =
+      sql"""SELECT DISTINCT kc.id, kc.googleProject
+            FROM KUBERNETES_CLUSTER AS kc
+            LEFT JOIN NODEPOOL AS np ON kc.id = np.clusterId
```
rtitle, line 76: I'm wondering if we could remove these LEFT JOINs and boil the interval checks into the `NOT EXISTS` subquery…

rtitle:
```
SELECT kc.id, kc.googleProject
FROM KUBERNETES_CLUSTER kc
WHERE
  kc.status != "DELETED" AND
  NOT EXISTS (
    SELECT *
    FROM NODEPOOL n
    LEFT JOIN APP a ON n.id = a.nodepoolId
    WHERE kc.id = n.clusterId AND 
      ((n.isDefault = 0 AND a.status != "DELETED" AND a.status != "ERROR") OR
        (app.destroyedDate > now() - INTERVAL 1 HOUR) OR
        (app.dateAccessed > now() INTERVAL 1 HOUR)));
```

rtitle: Does that work? This way the subquery checks for apps that are active OR have been recently deleted/accessed

resource-validator/src/main/resources/reference.conf
```diff
@@ -7,3 +7,8 @@ database {
 }
 
 report-destination-bucket = ${REPORT_DESTINATION_BUCKET}
+
+leonardo-pubsub {
+  google-project = "broad-dsde-dev"
```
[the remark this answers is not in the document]
rtitle: We will need to template this - I think a dedicated EV makes sense instead of parsing user creds

rtitle:
Overall looks good, 1 comment about simplifying the SQL query

core/src/main/scala/com/broadinstitute/dsp/models.scala
kyuksel: Huh...I thought IntelliJ implicit hints had pointed out those implicits were depended on by `Stream` piping 🤷‍♂️  Thanks for the catch.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
kyuksel: Great idea! I think it'll work. I just want to test it a bit more. My dev DB got into a weird state that I need to fix first.

kyuksel: Slightly modified version of that query (note the updates within the inner `WHERE` clause below) seems to work.

```sql
            SELECT kc.id, kc.googleProject
            FROM KUBERNETES_CLUSTER kc
            WHERE
              kc.status != "DELETED" AND
              NOT EXISTS (
                SELECT *
                FROM NODEPOOL np
                LEFT JOIN APP a ON np.id = a.nodepoolId
                WHERE
                  kc.id = np.clusterId AND np.isDefault = 0 AND
                  (
                    (a.status != "DELETED" AND a.status != "ERROR") OR
                    (a.status = "DELETED" AND a.destroyedDate > now() - INTERVAL 1 HOUR) OR
                    (a.status = "ERROR" AND a.dateAccessed > now() - INTERVAL 1 HOUR)
                  ));
```

Also want to do unit testing...

kyuksel:
@Qi77Qi @rtitle Thank you for your feedback! I think I've addressed them. Would you be able to re-review while I'm adding unit tests?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
```diff
@@ -44,22 +50,26 @@ object ResourceValidator {
...
         Stream.eval(InitBucketChecker.impl(deps.dbReader, checkRunnerDep).run(isDryRun))
       else Stream.empty
 
-      processes = Stream(deleteRuntimeCheckerProcess,
-                         errorRuntimeCheckerProcess,
-                         removeStagingBucketProcess,
-                         removeKubernetesClusters,
-                         deleteDiskCheckerProcess,
-                         removeInitBuckets).covary[F] //TODO: add more check
+      processes = Stream(
```
Qi77Qi, line 62: this needs to be reverted before merging

Qi77Qi (commented):
lgtm so far 👍 assuming you're still working on unit tests

resource-validator/src/main/resources/reference.conf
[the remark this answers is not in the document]
Qi77Qi: this needs to be address before merging

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -36,6 +35,37 @@ object DbReader {
...
+                SELECT *
+                FROM NODEPOOL np
+                LEFT JOIN APP a ON np.id = a.nodepoolId
+                WHERE
+                  kc.id = np.clusterId AND np.isDefault = 0 AND
+                  (
+                    (a.status != "DELETED" AND a.status != "ERROR") OR
+                    (a.status = "DELETED" AND a.destroyedDate > now() - INTERVAL 1 HOUR) OR
+                    (a.status = "ERROR" AND a.dateAccessed > now() - INTERVAL 1 HOUR) OR
+                    (a.id IS NULL)
```
kyuksel, line 62: @rtitle  ^^^ seems to take care of the exclusion of clusters with batch-pre-created nodepools. Added a [unit test](https://github.com/broadinstitute/leonardo-cron-jobs/pull/15/files#diff-d4eec159732c276b41661daac0de09192e20d7d261664c25dbc61d77bf0bb058R216) for that scenario as well.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -36,6 +35,37 @@ object DbReader {
...
+              NOT EXISTS (
+                SELECT *
+                FROM NODEPOOL np
+                LEFT JOIN APP a ON np.id = a.nodepoolId
+                WHERE
+                  kc.id = np.clusterId AND np.isDefault = 0 AND
+                  (
+                    (a.status != "DELETED" AND a.status != "ERROR") OR
+                    (a.status = "DELETED" AND a.destroyedDate > now() - INTERVAL 1 HOUR) OR
+                    (a.status = "ERROR" AND a.dateAccessed > now() - INTERVAL 1 HOUR) OR
```
Qi77Qi, line 61: wouldn't `a.status = "ERROR"` alone suffice? (without checkint `dateAccessed`)

rtitle: The `dateAccessed` check gives a grace period I think -- so we won't delete the cluster if an Error'd app is <1 hour old. 

I think I like this behavior, as a user might want to quickly recreate if they get a creation error.

rtitle:
The SQL looks good to me; I didn't review all the Scala in detail.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
Qi77Qi: > we won't delete the cluster if an Error'd app is <1 hour old

If we're trying to check if the cluster is < 1hr old, shouldn't we be checking `createdDate`? still not sure why `dateAccessed` is needed

kyuksel: Not the cluster but the app. Imagine a scenario where a user attempts to create an app but it errors out. In that case, we don't want to remove the cluster right away and instead, give the user some grace period in case they want to delete their app, create another app, etc.

kyuksel: Hmmm...on another look, we don't seem to update `app.dateAccessed` upon app creation erroring out. In that case this query would be effectively checking `createdDate`.

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/mocks.scala
```diff
@@ -12,9 +14,18 @@ class FakeDbReader extends DbReader[IO] {
...
 
-  override def getK8sClustersToDelete: Stream[IO, GKEModels.KubernetesClusterId] = Stream.empty
+  override def getKubernetesClustersToDelete: Stream[IO, KubernetesClusterToRemove] = Stream.empty
 
   override def getDeletedDisks: Stream[IO, Disk] = Stream.empty
 
   override def getInitBucketsToDelete: Stream[IO, InitBucketToRemove] = Stream.empty
 }
+
+class FakeGooglePublisher extends GooglePublisher[IO] {
```
Qi77Qi, line 24: not necessarily blocking this PR..but can you add this to wb-libs

kyuksel: yea - I can add as a follow-up PR - there's something else I'd like to add to wb-libs as well

core/src/main/scala/com/broadinstitute/dsp/models.scala
```diff
@@ -30,3 +31,9 @@ final case class InitBucketToRemove(googleProject: GoogleProject, bucket: Option
 
 final case class K8sClusterToScan(id: Long, kubernetesClusterId: KubernetesClusterId)
 final case class NodepoolToScan(id: Long, nodepoolId: NodepoolId)
+
+final case class KubernetesClusterToRemove(id: Long, googleProject: GoogleProject)
+// TODO: 'project' below is unnecessary but removing it requires an accompanying change in back Leo so leaving for now
+final case class DeleteKubernetesClusterMessage(clusterId: Long, project: GoogleProject, traceId: Option[TraceId]) {
```
Qi77Qi, line 37: minor, can you make this `TraceId` non optional. also I would probably move this to `KubernetesClusterRemover` file instead of puttng it in shared model file

Qi77Qi (approved):

rtitle (approved):

rtitle:
Btw, I ran this in prod this morning after yesterday's release:
```
MySQL [leonardo]> SELECT kc.id, kc.googleProject, kc.creator
    ->             FROM KUBERNETES_CLUSTER kc
    ->             WHERE
    ->               kc.status != "DELETED" AND
    ->               NOT EXISTS (
    ->                 SELECT *
    ->                 FROM NODEPOOL np
    ->                 LEFT JOIN APP a ON np.id = a.nodepoolId
    ->                 WHERE
    ->                   kc.id = np.clusterId AND np.isDefault = 0 AND
    ->                   (
    ->                     (a.status != "DELETED" AND a.status != "ERROR") OR
    ->                     (a.status = "DELETED" AND a.destroyedDate > now() - INTERVAL 1 HOUR) OR
    ->                     (a.status = "ERROR" AND a.createdDate > now() - INTERVAL 1 HOUR) OR
    ->                     (a.id IS NULL)
    ->                   )
    ->                 );
+----+------------------------+--------------------------+
| id | googleProject          | creator                  |
+----+------------------------+--------------------------+
|  2 | mschatz-jamboree       | michael.schatz@gmail.com |
|  3 | anvil-tool-development | michael.schatz@gmail.com |
|  5 | ea-anvil               | afgane@gmail.com         |
+----+------------------------+--------------------------+
3 rows in set (0.054 sec)
```

PR #17, opened 2020-10-20
[no ticket] remove report-destination-bucket from being populated from environment variable

Qi77Qi:
`report-destination-bucket` will be put in application.conf in https://github.com/broadinstitute/terra-helm/pull/165


kyuksel (approved):
👍

PR #18, opened 2020-10-20
[no ticket] update readme

Qi77Qi:


resource-validator/src/main/resources/application.conf.example
```diff
@@ -0,0 +1,3 @@
+leonardo-pubsub.google-project = "broad-dsde-dev" # or project you want to test against
+
+report-destination-bucket = "<YOUR-TESTING-BUCKET>" # if you're using leonardo's dev SA for local testing, make sure to give leonardo's dev SA permission to this bucket
```
kyuksel, line 3: Perhaps specify what permission to give?

kyuksel (approved):
👍 with one potential improvement inline

PR #19, opened 2020-10-23
IA-2297 restructure configs a bit

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2297

Different cron jobs can have difference configurations. I started having the same config for all jobs, but as cron jobs config starts to diverge, this doesn't work very well anymore.

This PR wil make each cron job reads their own config.


core/src/main/scala/com/broadinstitute/dsp/package.scala
```diff
@@ -11,10 +13,10 @@ package object dsp {
   val regionName = RegionName("us-central1")
 
   def initGoogleCredentials[F[_]: Sync: Timer](
```
kyuksel, line 15: Is this method used anywhere?

core/src/main/scala/com/broadinstitute/dsp/checkerDeps.scala
```diff
@@ -13,13 +15,13 @@ import org.broadinstitute.dsde.workbench.google2.{
...
 }
-import org.broadinstitute.dsde.workbench.model.google.GoogleProject
+import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GoogleProject}
 
 import scala.jdk.CollectionConverters._
 
 object RuntimeCheckerDeps {
   def init[F[_]: Concurrent: ContextShift: StructuredLogger: Parallel: Timer](
-    appConfig: AppConfig,
+    appConfig: RuntimeCheckerConfig,
```
kyuksel, line 24: Perhaps update the variable name too?

resource-validator/src/test/resources/reference.conf
```diff
@@ -6,4 +6,13 @@ database {
   password = "leonardo-test"
 }
 
-report-destination-bucket = "qi-test"
\ No newline at end of file
+report-destination-bucket = "qi-test"
```
kyuksel, line 9: perhaps rename this as `test-bucket` or something?

kyuksel (approved):
LGTM 👍   with minor comments inline

core/src/main/scala/com/broadinstitute/dsp/package.scala
Qi77Qi: made it used for one site...we can potentially use it in more places

PR #24, opened 2020-10-27
IA-2312 Add metrics for each check type

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2312

Related PR https://github.com/DataBiosphere/leonardo/pull/1670


core/src/main/scala/com/broadinstitute/dsp/CheckRunner.scala
```diff
@@ -38,6 +39,7 @@ trait CheckRunner[F[_], A] {
         .parEvalMapUnordered(50)(rt => checkResource(rt, isDryRun).handleErrorWith(_ => F.pure(None)))
         .unNone
         .map(_.toString)
+        .evalTap(_ => dependencies.metrics.incrementCounter(s"${appName}/${configs.checkType}"))
```
Qi77Qi, line 42: `evalTap` is similar to `observe`, it performs some effects but doesn't change the original Stream

rtitle (approved):
👍 great to add metrics, we can add dashboards once this is in

PR #29, opened 2020-11-05
IA-2331 update deletedFrom

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2331


kyuksel (approved):
LGTM 👍  My two cents would be to rename `deletedFrom` as `deletedBy` as the latter sounds more idiomatic to me.

Qi77Qi:
@kyuksel yea..good idea....but it's probably more confusing if we call it `deletedBy` here but `deletedFrom` in actual database and in leonardo....

kyuksel:
> it's probably more confusing if we call it deletedBy here but deletedFrom in actual database and in Leonardo....

For certain...I had meant we'd rename the column in Leo DB as well.

Qi77Qi:
Sent a discussion in slack...doesn't seem like there's a strong preference. so going to merge the PR as is

PR #31, opened 2020-11-10
[IA-2305] Refactor some code to wb-libs

jdcanas:
- Uses new version of cats.mtl which change `ApplicativeAsk` -> `Ask` 
- Uses new wb-libs function publishOne instead of the former streaming approach
- Refactors a mock and decoder to wb-libs


Qi77Qi (approved):
👍  thank you!

Qi77Qi:
oh...tests need to be fixed

kyuksel (approved):
👍 
Nice to see unnecessary and/or duplicate code go away 😄

Qi77Qi:
@jdcanas think this can be merged now?

PR #35, opened 2020-12-01
[IA-2271] Check that dataproc workers in google match db

jdcanas:


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -0,0 +1,67 @@
...
+
+      override def resourceToScan: fs2.Stream[F, RuntimeWithWorkers] = dbReader.getRuntimesWithWorkers
+
+      override def checkResource(runtime: RuntimeWithWorkers, isDryRun: Boolean)(
+        implicit ev: Ask[F, TraceId]
+      ): F[Option[RuntimeWithWorkers]] =
+        for {
+          clusterOpt <- deps.dataprocService
+            .getCluster(runtime.r.googleProject, regionName, DataprocClusterName(runtime.r.runtimeName))
+          runtime <- clusterOpt.fold[F[Option[RuntimeWithWorkers]]](F.pure(None)) { c =>
```
jdcanas, line 39: Is there a better way to do this?

I don't think `traverse` or `traverse_` quite work because we want to go from Option[A] to F[Option[B]].

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -0,0 +1,67 @@
...
+              runtime.workerConfig.numberOfPreemptibleWorkers == c.getConfig.getSecondaryWorkerConfig.getNumInstances
+            val isAnomalyDetected = !(doesPrimaryWorkerMatch && doesSecondaryWorkerMatch)
+
+            isAnomalyDetected match {
+              case true =>
+                val log =
+                  if (isDryRun)
+                    logger
+                      .warn(
+                        s"${runtime} has an anomaly with the number of workers in google. \n\tPrimary work match status: ${doesPrimaryWorkerMatch}\n\tSecondary worker match status: ${doesSecondaryWorkerMatch}"
```
Qi77Qi, line 52: primary worker?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -0,0 +1,67 @@
...
+                      )
+                  else
+                    // we want to see if this occurs before doing anything. Initial scans show 0 occurrences in dev.
+                    logger.warn(
+                      s"${runtime} has an anomaly with the number of workers in google. \n\tPrimary work match status: ${doesPrimaryWorkerMatch}\n\tSecondary worker match status: ${doesSecondaryWorkerMatch}"
```
Qi77Qi, line 57: shall we resize the cluster if not in `dryRun` mode?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
```diff
@@ -80,10 +85,11 @@ object ResourceValidator {
         removeInitBuckets,
         removeKubernetesClusters,
         deleteKubernetesClusterCheckerProcess,
-        deleteNodepoolCheckerProcess
+        deleteNodepoolCheckerProcess,
+        workerProcess
       ).covary[F]
 
-      _ <- processes.parJoin(8) // Update this number as we add more streams
+      _ <- processes.parJoin(9) // Update this number as we add more streams
```
Qi77Qi, line 92: maybe just use `parJoinUnbounded`? I think it'd be fine

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -0,0 +1,67 @@
...
+
+object DataprocWorkerChecker {
+  def impl[F[_]: Timer](
+    dbReader: DbReader[F],
+    deps: RuntimeCheckerDeps[F]
+  )(implicit F: Concurrent[F], logger: Logger[F], ev: Ask[F, TraceId]): CheckRunner[F, RuntimeWithWorkers] =
+    new CheckRunner[F, RuntimeWithWorkers] {
+      override def appName: String = resourceValidator.appName
+
+      override def configs = CheckRunnerConfigs(s"dataproc-workers", shouldAlert = true)
```
Qi77Qi, line 27: how about `number-of-dataproc-workers`? just be a bit more specific

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
jdcanas: it can be both, thats why i log and record both

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
jdcanas: I don't think we should  do to all the limitations around worker number changing  (i.e. if its the primary workers that don't match, sometimes the cluster needs to be deleted/re-created. This seems problematic if a user is working). I think we should see if this actually happens in the wild before investing the time to do this, as it would most likely require a pubsub message and adding the delete -> update -> re-create flow to leo similar to the stop -> update -. start flow we have. Even then, it could  be surprising for users.

I think the best course of action is to see if it actually happens, and then we can do the easy re-size of secondary  workers and layer in the primary worker update messaging

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
Qi77Qi: I meant you have a typo `Primary work match status`

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
Qi77Qi: @kyuksel since you're the expert on resizing dataproc cluster now, what do you think?

kyuksel: > I don't think we should do to all the limitations around worker number changing (i.e. if its the primary workers that don't match, sometimes the cluster needs to be deleted/re-created.

I wasn't aware of this limitation since for my work, I've only been concerned with resizing to remove preemptibles. I didn't see such a limitation be mentioned in the [docs](https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/scaling-clusters#:~:text=After%20creating%20a%20Dataproc%20cluster,Use%20Dataproc%20Autoscaling.) either. Let me play around quickly with it and report back.

> I think we should see if this actually happens in the wild before investing the time to do this

I can see how that can make sense. What would be the checkpoint mechanism to re-evaluate where things stand on this? The concern being we may forget...

Also a tangential question: I wonder if we should only take into consideration Dataproc clusters in `Running` status, especially if we decide to take action on anomalies.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -53,27 +46,37 @@ object DataprocWorkerChecker {
 
             isAnomalyDetected match {
```
kyuksel, line 47: Nit: `if (isAnomalyDetected) {...}` may be simpler

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -53,27 +46,37 @@ object DataprocWorkerChecker {
...
-                        .resizeCluster(runtime.r.googleProject,
-                          regionName,
-                          DataprocClusterName(runtime.r.runtimeName),
+                if (isDryRun)
+                  logger
+                    .warn(
+                      s"${runtime} has an anomaly with the number of workers in google. \n\tPrimary worker match status: $doesPrimaryWorkerMatch\n\tSecondary worker match status: ${doesSecondaryWorkerMatch}"
+                    )
+                    .as[Option[RuntimeWithWorkers]](Some(runtime))
+                // If the number of primary workers is less than 2, modifying workers requires a creation and deletion. Leo does not handles these, so we will not either
```
kyuksel, line 55: Typo

```suggestion
                // If the number of primary workers is less than 2, modifying workers requires a creation and deletion. Leo does not handle these, so we will not either
```

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -53,27 +46,37 @@ object DataprocWorkerChecker {
...
-                .as[Option[RuntimeWithWorkers]](Some(runtime))
-                    else
-                      deps.checkRunnerDeps.metrics.incrementCounter(s"$appName/$unfixableAnomalyCheckType") >>
-                      logger.warn(s"${runtime} has an anomaly with the number of workers in google. Unable to fix the anomaly. Recording a metric and moving on")
-                        .as[Option[RuntimeWithWorkers]](Some(RuntimeWithWorkers(runtime.r, runtime.workerConfig, false)))
-                log
+                    ) >>
+                    logger
+                      .warn(
+                        s"${runtime} has an anomaly with the number of workers in google. \n\tPrimary work match status: $doesPrimaryWorkerMatch\n\tSecondary worker match status: ${doesSecondaryWorkerMatch}"
```
kyuksel, line 67:
```suggestion
                        s"${runtime} has an anomaly with the number of workers in google. \n\tPrimary worker match status: $doesPrimaryWorkerMatch\n\tSecondary worker match status: ${doesSecondaryWorkerMatch}"
```

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -0,0 +1,86 @@
...
+                      regionName,
+                      DataprocClusterName(runtime.r.runtimeName),
+                      if (doesPrimaryWorkerMatch) None else Some(runtime.workerConfig.numberOfWorkers),
+                      if (doesSecondaryWorkerMatch) None else Some(runtime.workerConfig.numberOfPreemptibleWorkers)
+                    ) >>
+                    logger
+                      .warn(
+                        s"${runtime} has an anomaly with the number of workers in google. \n\tPrimary work match status: $doesPrimaryWorkerMatch\n\tSecondary worker match status: ${doesSecondaryWorkerMatch}"
+                      )
+                      .as[Option[RuntimeWithWorkers]](Some(runtime))
```
kyuksel, line 69: Could be simplified as

```scala
.as(Option(runtime))
```

Similarly below...

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
kyuksel: `flatTraverse` should work similarly to in [this case](https://github.com/broadinstitute/leonardo-cron-jobs/blob/a89277b3d69835b27c5137a7c5f6391905254a8b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala#L67) unless I'm missing something.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -130,6 +132,14 @@ object DbReader {
          """
       .query[KubernetesClusterToRemove]
 
+  val dataprocClusterWithWorkersQuery =
+    sql"""SELECT DISTINCT c1.id, googleProject, clusterName, rt.cloudService, c1.status, rt.numberOfWorkers, rt.numberOfPreemptibleWorkers
+          FROM CLUSTER AS c1
+          INNER JOIN RUNTIME_CONFIG AS rt ON c1.`runtimeConfigId`=rt.id
+          WHERE rt.cloudService="DATAPROC" AND NOT rt.numberOfWorkers=0 AND rt.numberOfWorkers IS NOT NULL AND NOT c1.status="DELETED"
```
kyuksel, line 139: Do we catch anomalies where Leo DB indicates zero workers but there are non-zero workers in Google?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -0,0 +1,86 @@
...
+                    .as[Option[RuntimeWithWorkers]](Some(runtime))
+                // If the number of primary workers is less than 2, modifying workers requires a creation and deletion. Leo does not handles these, so we will not either
+                else if (c.getConfig.getWorkerConfig.getNumInstances >= 2)
+                  deps.dataprocService
+                    .resizeCluster(
+                      runtime.r.googleProject,
+                      regionName,
+                      DataprocClusterName(runtime.r.runtimeName),
+                      if (doesPrimaryWorkerMatch) None else Some(runtime.workerConfig.numberOfWorkers),
+                      if (doesSecondaryWorkerMatch) None else Some(runtime.workerConfig.numberOfPreemptibleWorkers)
```
kyuksel, line 63: Have we tested scenarios where the cluster is not in `RUNNING` status? The query below returns all non-`DELETED` clusters (with non-matching workers) and I'm not sure if trying to resize clusters in all those states would work.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/Main.scala
```diff
@@ -28,6 +28,9 @@ object Main
         val shouldCheckInitBuckets =
           Opts.flag("checkInitBuckets", "checks that init buckets for deleted runtimes are deleted").orFalse
 
+        val shouldCheckWorkers =
```
kyuksel, line 31: I wonder if we would want to be more descriptive as in `shouldCheckDataprocWorkers`.

core/src/test/scala/com/broadinstitute/dsp/Generators.scala
```diff
@@ -52,6 +52,12 @@ object Generators {
     googleProject <- genGoogleProject
   } yield KubernetesClusterToRemove(id, googleProject)
 
+  val genRuntimeWithWorkers: Gen[RuntimeWithWorkers] = for {
+    runtime <- genRuntime
+    num1 <- Gen.chooseNum(1, 100)
+    num2 <- Gen.chooseNum(1, 100)
```
kyuksel, line 58: Does it make sense to include `0` in the range as it could help test some corner cases in the `spec`?

kyuksel (changes requested):
Looks good overall. Comments inline.

core/src/test/scala/com/broadinstitute/dsp/Generators.scala
jdcanas: we do not include 0 here on purpose, since this data will never occur due to database filters. Additionally, it adds some annoying conditionals, where the second number must be 0 if the first is.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
jdcanas: I like match a bit better style-wise for large blocks, and if looks repetitive since there are so many ifs inside the match already

core/src/test/scala/com/broadinstitute/dsp/Generators.scala
kyuksel: What do you mean by _database filters_?

kyuksel: Did you mean the following line in the query

```
NOT rt.numberOfWorkers=0 AND rt.numberOfWorkers IS NOT NULL
```
?

Speaking of which, why are we filtering those out? Dataproc clusters can have only a master node and zero worker and secondary workers.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
kyuksel: I respect your taste 😊

Qi77Qi (commented):
lgtm..just curious, did you find any anomaly in dev?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
jdcanas: Hmm... I tested only on running clusters. It does error out on stopped clusters though. I disabled the update call for stopped clusters, but still record a metric and output a specific log.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -37,47 +38,63 @@ object DataprocWorkerChecker {
...
+                  case _ =>
+                    // Here we log a metric when we detect the aforementioned anomaly is unfixable (workers < 2). This metric is in addition to the one already reported for finding a worker mismatch
+                    deps.checkRunnerDeps.metrics.incrementCounter(s"$appName/$unfixableAnomalyCheckType") >>
+                      logger
+                        .warn(
+                          s"${runtime.toString} has an anomaly with the number of workers in google. Unable to fix the anomaly. Recording a metric and moving on."
+                        )
+                        .as(Option(runtime))
+                }
+              case false => F.pure[Option[RuntimeWithWorkers]](None)
```
kyuksel, line 97: [FYI] A slightly shorter way of writing that is

```scala
F.pure(none[RuntimeWithWorkers])
```

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -37,47 +38,63 @@ object DataprocWorkerChecker {
...
+                        } >>
+                        logger
+                          .warn(
+                            s"${runtime} has an anomaly with the number of workers in google. \n\tPrimary work match status: $doesPrimaryWorkerMatch\n\tSecondary worker match status: ${doesSecondaryWorkerMatch}"
+                          )
+                          .as(Option(runtime))
+                    else
+                      logger
+                        .warn(
+                          s"${runtime} has an anomaly with the number of workers in google, but we cannot attempt to fix it because the cluster is not running. \n\tprimary in google: ${c.getConfig.getWorkerConfig.getNumInstances} secondary in google ${c.getConfig.getSecondaryWorkerConfig.getNumInstances}."
```
kyuksel, line 85: Should this be
```suggestion
                          s"${runtime} has an anomaly with the number of workers in google, but we cannot attempt to fix it because the cluster is not running. \n\tNumber of primary workers in google: ${c.getConfig.getWorkerConfig.getNumInstances} Number of secondary workers in google: ${c.getConfig.getSecondaryWorkerConfig.getNumInstances}."
```
?

jdcanas:
> lgtm (less Kaan's comments)..just curious, did you find any anomaly in dev?

No,  I didn't. I created a cluster for all the various cases and modified the number of workers in the database to create  an anomaly to see logs/behavior

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
kyuksel: Let's make this change before merging.

kyuksel (approved):
👍🏼  with one missing log message change mentioned inline.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
jdcanas: Oops, i committed this but didn't push, thanks.

PR #36, opened 2020-12-14
[IA-2229] Resource Validator checks stopped runtimes

kyuksel:
`StoppedRuntimeChecker` and `stoppedRuntimeQuery` are the actual changes but I also couldn't help but make some non-functional changes to `Deleted` and `Errored` `RuntimeChecker`s as well as `Main` in hopes of making the code more consistent and readable stylistically. Sorry about the additional cognitive load due to that.


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
```diff
@@ -0,0 +1,106 @@
...
+            .emits(instances.values.toSeq.flatten)
+            .covary[F]
+            .parEvalMapUnordered(10) { instance =>
+              deps.computeService
+                .getInstance(project, zone, instance)
+                .handleErrorWith(_ => F.pure(None))
+            }
+            .exists(_.exists(_.getStatus == "RUNNING"))
+            .compile
+            .lastOrError
```
kyuksel, line 103: Is there a better way to go from `Stream[F, Boolean]` to `F[Boolean]` in this case?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedRuntimeChecker.scala
```diff
@@ -19,38 +18,38 @@ object DeletedRuntimeChecker {
     new CheckRunner[F, Runtime] {
       override def appName: String = resourceValidator.appName
 
-      override def configs = CheckRunnerConfigs(s"deleted-runtime", true)
+      override def configs = CheckRunnerConfigs(s"deleted-runtime", shouldAlert = true)
```
Qi77Qi, line 21: is it just some renaming and function moved around in this file?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeChecker.scala
```diff
@@ -47,7 +47,9 @@ object ErroredRuntimeChecker {
                   )
                   .as(Some(runtime))
               else
-                logger.warn(s"${runtime} still exists in ${cluster.getStatus.getState.name()} status. Going to delete") >> deps.dataprocService
+                logger.warn(
```
Qi77Qi, line 50: did you run scalafmt? I wouldn't format anything manually, and just leave formatting to scalafmt

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/Main.scala
```diff
@@ -16,43 +16,47 @@ object Main
...
               .run[IO](
-                isDryRun = dryRun,
-                shouldRunAll = runAll,
-                shouldRunCheckDeletedRuntimes = shouldCheckDeletedRuntimes,
-                shouldRunCheckErroredRuntimes = shouldRunCheckErroredRuntimes,
-                shouldRunCheckDeletedDisks = shouldRunCheckDeletedDisks,
-                shouldRunCheckInitBuckets = shouldRunCheckInitBuckets,
-                shouldRunCheckDeletedKubernetesCluster = shouldRunCheckDeletedKubernetesClusters,
-                shouldRunCheckDeletedNodepool = shouldRunCheckDeletedNodepools
+                dryRun,
```
Qi77Qi, line 51: we added the named param so that it's less easy to get the param order wrong....can you add it back

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
```diff
@@ -18,28 +18,33 @@ import org.broadinstitute.dsde.workbench.openTelemetry.OpenTelemetryMetrics
 object ResourceValidator {
   def run[F[_]: ConcurrentEffect: Parallel](isDryRun: Boolean,
                                             shouldRunAll: Boolean,
-                                            shouldRunCheckDeletedRuntimes: Boolean,
-                                            shouldRunCheckErroredRuntimes: Boolean,
-                                            shouldRunCheckDeletedDisks: Boolean,
-                                            shouldRunCheckInitBuckets: Boolean,
+                                            shouldCheckDeletedRuntimes: Boolean,
```
Qi77Qi, line 21: let's not re-order these params around much...they're all `Boolean`, which means if you get the order wrong, we can't know at compile time

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
Qi77Qi: `lastOrError` is the right one to use

Qi77Qi (commented):
the actual change looks good to me in `StoppedRuntimeChecker`...are you still going to add `StoppedRuntimeCheckerSpec.scala`? 

also questions.
1. did you run db tests?
2. have you run this locally against dev?

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeCheckerSpec.scala
```diff
@@ -33,18 +33,18 @@ class ErroredRuntimeCheckerSpec extends AnyFlatSpec with CronJobsTestSuite {
...
+      val res = erroredRuntimeChecker.checkResource(runtime, dryRun)
       res.unsafeRunSync() shouldBe None
     }
   }
 
   it should "return Runtime if runtime still exists in Google" in {
     forAll { (runtime: Runtime, dryRun: Boolean) =>
       val dbReader = new FakeDbReader {
-        override def getDeletedRuntimes: fs2.Stream[IO, Runtime] = Stream.emit(runtime)
+        override def getErroredRuntimes: fs2.Stream[IO, Runtime] = Stream.emit(runtime)
```
Qi77Qi, line 47: thanks for fixing these

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedRuntimeChecker.scala
kyuksel: Yes, no functional change.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeChecker.scala
kyuksel: Yes, I had run `scalafmtAll`. Will run again after addressing the PR feedback.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/Main.scala
kyuksel: I see. Sure.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
kyuksel: Understood. Part of my motivation was to reduce future errors too, by grouping them more logically.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
kyuksel: thanks

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeCheckerSpec.scala
kyuksel: 🖖🏼

kyuksel:
@Qi77Qi Thanks for the feedback. Yes, I will still add `StoppedRuntimeCheckerSpec`. I did run the DB tests and locally against dev. Will run for some more cases as well.

core/src/test/scala/com/broadinstitute/dsp/Generators.scala
```diff
@@ -11,7 +11,7 @@ object Generators {
     cloudService <- genCloudService
     project <- genGoogleProject
     runtimeName <- Gen.uuid.map(_.toString)
-    status <- Gen.oneOf("Running", "Creating", "Deleted", "Error")
+    status <- Gen.oneOf("RUNNING", "CREATING", "DELETED", "ERROR")
```
Qi77Qi, line 14: the origina generator is correct...`status` for `Runtime` is not all upper case

kyuksel:
@Qi77Qi I believe I've addressed all your comments. The diff since your last review is [here](https://github.com/broadinstitute/leonardo-cron-jobs/pull/36/files/84b6433a1fb0a899a37ad9cddcc6df2718ee9642..d9af1b1025055cdc7d0a568c262e4a3310bdf769).

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeCheckerSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+                                                                       InstanceName("secondary-worker-instance-2"))
+  )
+
+  def makeClusterWithStatus(status: String): Cluster = {
+    val validStatus = status match {
+      case "RUNNING"  => "RUNNING"
+      case "CREATING" => "CREATING"
+      case "ERROR"    => "ERROR"
+      case "DELETED"  => "DELETING"
+      case _          => "UNRECOGNIZED"
```
Qi77Qi, line 175: when will `_` happen?

core/src/test/scala/com/broadinstitute/dsp/Generators.scala
kyuksel: Ah I think I confused `Runtime` which models the Leo DB query result with `Instance` (from Google library) status which _is_ all upper case.

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeCheckerSpec.scala
kyuksel: That's to guard against a caller passing in some other string. It wouldn't happen if the caller uses `genRuntime` but we can't know that. `genRuntime` may also change in the future.

Qi77Qi: wouldn't that be a bug? since this is in test, I woudln't mind just throwing an exception if the input string is any invalid status (or making the function return `Option` or `Either` if you feel like).

kyuksel: I don't think it would necessarily be a bug but getting an exception in such a case sounds good.

Qi77Qi (approved):

PR #40, opened 2021-01-11
[IA-2352] add check if billing is enabled to checkers

jdcanas:
In non-dry runs, the deleted and errored runtime checkers first check if billing is enabled before attempted to delete the runtime. 

This relies on the merged wb-libs PR is here: https://github.com/broadinstitute/workbench-libs/pull/474/files

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
   val logbackVersion = "1.2.3"
-  val workbenchGoogle2Version = "0.18-b4202c9b-SNAP"
+  val workbenchGoogle2Version = "0.19-bc594f9"
   val doobieVersion = "0.9.4"
   val openTelemetryVersion = "0.1-e66171c"
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
   val logbackVersion = "1.2.3"
-  val workbenchGoogle2Version = "0.18-c9edd8e"
+  val workbenchGoogle2Version = "0.18-b4202c9b-SNAP"
   val doobieVersion = "0.9.4"
   val openTelemetryVersion = "0.1-e66171c"
 
@@ -23,9 +23,11 @@ object Dependencies {
     "com.google.cloud" % "google-cloud-dataproc" % "0.122.1",
     "com.google.cloud" % "google-cloud-compute" % "0.118.0-alpha",
     "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2Version,
+    "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2Version,
     "org.broadinstitute.dsde.workbench" %% "workbench-google2" % workbenchGoogle2Version % Test classifier "tests",
     "org.scalatestplus" %% "scalacheck-1-14" % "3.2.2.0" % Test,
-    "org.scalatestplus" %% "mockito-3-3" % "3.2.0.0" % Test // https://github.com/scalatest/scalatestplus-selenium
+    "org.scalatestplus" %% "mockito-3-3" % "3.2.0.0" % Test, // https://github.com/scalatest/scalatestplus-selenium
+    "ca.mrvisser" %% "sealerate" % "0.0.6"
   )
 
   val resourceValidator = core
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedRuntimeChecker.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedRuntimeChecker.scala
@@ -13,14 +13,14 @@ import org.broadinstitute.dsde.workbench.model.TraceId
 object DeletedRuntimeChecker {
   def impl[F[_]: Timer](
     dbReader: DbReader[F],
-    deps: RuntimeCheckerDeps[F]
+    deps: BillingDeps[F]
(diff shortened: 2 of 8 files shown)


project/Dependencies.scala
```diff
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
   val logbackVersion = "1.2.3"
-  val workbenchGoogle2Version = "0.18-c9edd8e"
+  val workbenchGoogle2Version = "0.18-b4202c9b-SNAP"
```
jdcanas, line 5: TODO: change once wb-libs PR is merged

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedRuntimeChecker.scala
```diff
@@ -37,28 +37,41 @@ object DeletedRuntimeChecker {
         implicit ev: Ask[F, TraceId]
       ): F[Option[Runtime]] =
         for {
-          clusterOpt <- deps.dataprocService
+          clusterOpt <- deps.runtimeCheckerDeps.dataprocService
             .getCluster(runtime.googleProject, regionName, DataprocClusterName(runtime.runtimeName))
-          _ <- clusterOpt.traverse_ { _ =>
+          isBillingEnabled <- deps.billingService.isBillingEnabled(runtime.googleProject)
```
Qi77Qi, line 42: we only need to check `isBillingEnabled ` if cluster still exists in google..can we move this inside `clusterOpt` just that we can save this call in most cases

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedRuntimeChecker.scala
```diff
@@ -37,28 +37,42 @@ object DeletedRuntimeChecker {
...
+                  .as(Option(runtime))
+              else
+                isBillingEnabled match {
+                  case true =>
+                    logger.warn(s"$runtime still exists in Google and billing is enabled. Going to delete it.") >> deps.runtimeCheckerDeps.dataprocService
+                      .deleteCluster(runtime.googleProject, regionName, DataprocClusterName(runtime.runtimeName))
+                      .as(Option(runtime))
+                  case false =>
+                    logger
+                      .warn(s"$runtime still exists with an anomaly, but cannot delete it because billing is disabled.")
```
Qi77Qi, line 59: shall we log it at `info` level, and just say something like "$runtime doesn't exist in google anymore because billing is disable for $project"

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeChecker.scala
```diff
@@ -32,8 +32,9 @@ object ErroredRuntimeChecker {
 
       def checkDataprocCluster(runtime: Runtime, isDryRun: Boolean)(implicit ev: Ask[F, TraceId]): F[Option[Runtime]] =
         for {
-          clusterOpt <- deps.dataprocService
+          clusterOpt <- deps.runtimeCheckerDeps.dataprocService
             .getCluster(runtime.googleProject, regionName, DataprocClusterName(runtime.runtimeName))
+          isBillingEnabled <- deps.billingService.isBillingEnabled(runtime.googleProject)
```
Qi77Qi, line 37: thanks for addressing ErroredRuntime as well!

Same comment for moving isBillingEnabled check down

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedRuntimeChecker.scala
jdcanas: It does exist in google though, we just cannot delete it because billing is disabled.

Qi77Qi: my guess is the cluster doesn't actually exist in google (it's just getCluster call returns a Some) because for it to exist, it'll incur cost to google, and google can't charge for it

jdcanas: i changed it to `$runtime has been reported from getCluster, but billing is disabled so cannot perform any actions`

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeChecker.scala
```diff
@@ -13,12 +13,12 @@ import org.broadinstitute.dsde.workbench.model.TraceId
 object ErroredRuntimeChecker {
   def iml[F[_]: Timer](
     dbReader: DbReader[F],
-    deps: RuntimeCheckerDeps[F]
+    deps: BillingDeps[F]
```
Qi77Qi, line 16: just curious, any reason why cloudBilling wasn't added to `RuntimeCheckerDeps` directly?

Qi77Qi (approved):

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ErroredRuntimeChecker.scala
jdcanas: It could have been. I did it this way originally to separate out the google1 dep, but thats no longer relevant. I like the idea of keeping runtimeCheckerDeps concise here though, its a pretty specific dep I don't think much else in the resource validator will care about

PR #41, opened 2021-01-14
IA-2458 if billing is disabled, update Error cause properly

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2458

1. If cluster is in Error due to billing disabled, update this in DB
2. Add `deleteFrom` to `zombie cron job` if runtime is marked as Deleted from zombie job

### DIFF
--- a/core/src/main/scala/com/broadinstitute/dsp/checkerDeps.scala
+++ b/core/src/main/scala/com/broadinstitute/dsp/checkerDeps.scala
@@ -41,9 +41,10 @@ object RuntimeCheckerDeps {
                                                               blocker,
                                                               regionName,
                                                               blockerBound)
+      billingService <- GoogleBillingService.fromCredential(scopedCredential, blocker, blockerBound)
     } yield {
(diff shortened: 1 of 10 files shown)


core/src/main/scala/com/broadinstitute/dsp/checkerDeps.scala
```diff
@@ -76,5 +78,3 @@ final case class NodepoolCheckerDeps[F[_]](checkRunnerDeps: CheckRunnerDeps[F],
 final case class DiskCheckerDeps[F[_]](checkRunnerDeps: CheckRunnerDeps[F], googleDiskService: GoogleDiskService[F])
 
 final case class RuntimeCheckerConfig(pathToCredential: Path, reportDestinationBucket: GcsBucketName)
-
```
Qi77Qi, line 79: @jdcanas I removed this since zombie job also needs billingService now

jdcanas (approved):

PR #43, opened 2021-01-20
IA-2415 update query to only care about recently deleted runtimes

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2415

only scan runtimes, disks that are deleted in the last 90 days. Assuming that runtimes/disks deleted earlier then 90 days have already been checked many times

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
@@ -27,13 +27,16 @@ object DbReader {
 
   val deletedDisksQuery =
     sql"""
-           select pd1.id, pd1.googleProject, pd1.name from PERSISTENT_DISK as pd1 where pd1.status="Deleted" and
-           NOT EXISTS
-           (
-             SELECT *
-             FROM PERSISTENT_DISK pd2
-             WHERE pd1.googleProject = pd2.googleProject and pd1.name = pd2.name and pd2.status != "Deleted"
-          )
+           select pd1.id, pd1.googleProject, pd1.name 
+           FROM PERSISTENT_DISK as pd1 
+           WHERE pd1.status="Deleted" and
+             pd1.destroyedDate > now() - INTERVAL 90 DAY AND
+             NOT EXISTS
+             (
+               SELECT *
+               FROM PERSISTENT_DISK pd2
+               WHERE pd1.googleProject = pd2.googleProject and pd1.name = pd2.name and pd2.status != "Deleted"
+              )
         """.query[Disk]
 
   val initBucketsToDeleteQuery =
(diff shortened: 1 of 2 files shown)


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -27,13 +27,16 @@ object DbReader {
...
-           select pd1.id, pd1.googleProject, pd1.name from PERSISTENT_DISK as pd1 where pd1.status="Deleted" and
-           NOT EXISTS
-           (
-             SELECT *
-             FROM PERSISTENT_DISK pd2
-             WHERE pd1.googleProject = pd2.googleProject and pd1.name = pd2.name and pd2.status != "Deleted"
-          )
+           select pd1.id, pd1.googleProject, pd1.name 
+           FROM PERSISTENT_DISK as pd1 
+           WHERE pd1.status="Deleted" and
```
kyuksel, line 32: Nit:
```suggestion
           FROM PERSISTENT_DISK AS pd1 
           WHERE pd1.status="Deleted" AND
```

jdcanas, line 32: small nit, but typically its good to put the linking conditional at the beginning of the next line in long queries, I think it read a bit better (i.e. move the `AND` thats above + below to the beginning of the subsequent line

kyuksel (commented):
The changes look good to me. One thing I noticed...the ticket and the PR description say to include runtimes but the changes only touch disks and buckets. Was there going to be a follow-up PR?

jdcanas (approved):

Qi77Qi:
@kyuksel buckets are not touched...I thikn you might read it wrong becuz the bucket line was above the diff

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
kyuksel: Ha..interesting..I have the opposite preference (i.e. having `AND` at the end like it is above). I understand that it's a personal taste and wouldn't push back against. The one thing I do feel a bit strongly about is consistency between queries, though. Would love to pick one style and apply it to all if possible 😊

kyuksel:
> buckets are not touched...I thikn you might read it wrong because the bucket line was above the diff

Oh yea…ha…I totally got tripped up by the diff collapsing the code  🤦🏼

kyuksel (approved):

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
Qi77Qi: I like having `AND` at the end as well...going to leave things as is since it's just personal preference

PR #44, opened 2021-02-04
[no ticket] Attempt to make db connection error less often

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2490

```
Exception in thread "main" java.sql.SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 30000ms.
        at com.zaxxer.hikari.pool.HikariPool.createTimeoutException(HikariPool.java:695)
        at com.zaxxer.hikari.pool.HikariPool.getConnection(HikariPool.java:197)
```

This error seems to happen more frequently recently in resource-validator. I asked in doobie gitter channel, the suggestion is read all records into memory instead because `.stream.transact(xa)` only closes connection once the `Stream` terminates.

I'm thinking we can reduce the number of records even further by reducing number of days we look back. Since resource-validator runs every day, so technically, this should totally be fine.

Another change I made is reduce parallelism for `resource-validator` in hope of reducing the connection not available error

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
@@ -97,7 +97,7 @@ object ResourceValidator {
         workerProcess
       ).covary[F]
 
-      _ <- processes.parJoinUnbounded
+      _ <- processes.parJoin(9)
     } yield ExitCode.Success
   }.drain
 
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,9 +2,9 @@ import sbt._
 
 object Dependencies {
   val logbackVersion = "1.2.3"
-  val workbenchGoogle2Version = "0.19-f0578d6"
+  val workbenchGoogle2Version = "0.19-1aba7fd"
   val doobieVersion = "0.10.0"
-  val openTelemetryVersion = "0.1-f0578d6"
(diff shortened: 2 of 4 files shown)


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
```diff
@@ -97,7 +97,7 @@ object ResourceValidator {
         workerProcess
       ).covary[F]
 
-      _ <- processes.parJoinUnbounded
+      _ <- processes.parJoin(9)
```
rtitle, line 100: Seems there are 11 sub-streams.. should this be parJoin(11)?

rtitle (approved):
The change looks fine. Also makes sense though that the DB connection is not closed until the stream terminates. It might be better to read DB data into memory; close the connection; and only then make Google API calls, etc.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/ResourceValidator.scala
Qi77Qi: yea, I intentionally made this 9, hoping reducing the number of Streams running in parallel will make the connection error less often

Qi77Qi:
@rtitle yea, I think that'll be the next thing to do if this doesn't help much...but I wanna see if this is enough to make things more stable first :)

kyuksel (commented):
👍🏼 🤞🏼

PR #45, opened 2021-02-05
[IA-2500] don't report anomaly if preemptibles in google is 0

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2500

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
@@ -40,13 +40,14 @@ object DataprocWorkerChecker {
             .getCluster(runtime.r.googleProject, regionName, DataprocClusterName(runtime.r.runtimeName))
           runtime <- clusterOpt.flatTraverse { c =>
             val doesPrimaryWorkerMatch =
-              if (c.getConfig.getWorkerConfig.getNumInstances == 0)
+              runtime.workerConfig.numberOfWorkers.getOrElse(0) == c.getConfig.getWorkerConfig.getNumInstances
+            val doesSecondaryWorkerMatch =
+              if (c.getConfig.getSecondaryWorkerConfig.getNumInstances == 0 && runtime.r.status.toLowerCase == "stopped")
                 true //pre-emptible workers can disappear, which isn't an anomaly if it doesn't match what we have in Leo DB
               else
-                runtime.workerConfig.numberOfWorkers.getOrElse(0) == c.getConfig.getWorkerConfig.getNumInstances
-            val doesSecondaryWorkerMatch =
-              runtime.workerConfig.numberOfPreemptibleWorkers
-                .getOrElse(0) == c.getConfig.getSecondaryWorkerConfig.getNumInstances
+                runtime.workerConfig.numberOfPreemptibleWorkers
+                  .getOrElse(0) == c.getConfig.getSecondaryWorkerConfig.getNumInstances
+
             val isAnomalyDetected = !(doesPrimaryWorkerMatch && doesSecondaryWorkerMatch)
 
             isAnomalyDetected match {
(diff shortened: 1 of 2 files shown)


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -40,7 +40,10 @@ object DataprocWorkerChecker {
             .getCluster(runtime.r.googleProject, regionName, DataprocClusterName(runtime.r.runtimeName))
           runtime <- clusterOpt.flatTraverse { c =>
             val doesPrimaryWorkerMatch =
-              runtime.workerConfig.numberOfWorkers.getOrElse(0) == c.getConfig.getWorkerConfig.getNumInstances
+              if (c.getConfig.getWorkerConfig.getNumInstances == 0)
```
kyuksel, line 43: we need to add to the if-clause 

```scala
runtime.r.status.toLowerCase == "stopped"
```

since we want to ignore the zero preemptible case only if the cluster was stopped because we remove preemptibles before stopping the cluster

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
```diff
@@ -40,7 +40,10 @@ object DataprocWorkerChecker {
...
+                true //pre-emptible workers can disappear, which isn't an anomaly if it doesn't match what we have in Leo DB
```
kyuksel, line 44: It'd be good to modify the comment a bit to the effect of:

```scala
// We remove preemptibles before stopping clusters. Therefore, we don't consider this case an anomaly.
```

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
kyuksel: Also, the check above needs to go into the 

```scala
val doesSecondaryWorkerMatch = ...
```

block instead, no? Preemptibles are secondary workers.

kyuksel (changes requested):

kyuksel:
@Qi77Qi Some of the comments I made inline are with the assumption that the case that's been causing false anomalies is when a Dataproc cluster is stopped and therefore has 0 preemptibles, rather than the cluster is running but Google has re-claimed the preemptibles. Unless we know the latter has been causing false anomalies, I'd keep that check in as I doubt it's something that'll cause false alarms much.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
Qi77Qi: yep...this actually caused unit tests...posted the PR too soon :)

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala
Qi77Qi: I see..I didn't know we remove preemptiles on our end

Qi77Qi:
@kyuksel think I address comments, also added a unit test. Let me know if there's any other concerns

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerCheckerSpec.scala
```diff
@@ -187,4 +191,46 @@ class DataprocWorkerCheckerSpec extends AnyFlatSpec with CronJobsTestSuite {
       res.unsafeRunSync() shouldBe Some(runtime)
     }
   }
+
+  it should "return None if cluster in google secondary worker numbers don't match the db" in {
```
kyuksel, line 195:
```suggestion
  it should "return None if preemptible worker numbers don't match when a cluster is stopped in Google and has zero preemptibles" in {
```

kyuksel (approved):
Looks good 👍🏼 Minor suggestion inline. Thanks for adding a test too.

PR #46, opened 2021-02-26
Unlink disk when deleting runtime and apps [IA-2524]

ehanna4:
Previously when the cron-job ran to delete a runtime or app it wasn't unlinking the disks attached to those deleted resources resulting in errors for users trying to make new resources since the disks were seen as attached. 

(also cleaned up a unit test around stopAfterCreation which is a removed property)

Added unit tests and tested running the cron-job locally

Steps to reproduce to test:
1. Create a new GCE Cluster in dev
2. Delete the VM through google console
3. Run Zombie-Monitor (had to change the 1hr wait period)
4. Check the DB

Before with a cluster in `Running` and successfully deleting the VM in google
Screen Shot 2021-03-02 at 10 15 21 AM

After running the cron-job the PD is set to `NULL` for that runtime
Screen Shot 2021-03-02 at 10 17 23 AM

Tested the K8'sCluster version making a Galaxy app and deleting the cluster in google and checking the DB before and after running zombie-monitor
Screen Shot 2021-03-02 at 3 26 57 PM


core/src/test/scala/com/broadinstitute/dsp/DBTestHelper.scala
```diff
@@ -58,54 +58,52 @@ object DBTestHelper {
...
-         stagingBucket, 
-         dateAccessed, 
-         autopauseThreshold, 
-         defaultClientId, 
-         stopAfterCreation, 
-         kernelFoundBusyDate, 
-         welderEnabled, 
-         internalId, 
-         runtimeConfigId, 
+         (clusterName,
```
Qi77Qi, line 61: a bit hard to see from the diff..but what's changed in this list?

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedKubernetesClusterChecker.scala
```diff
@@ -33,7 +33,7 @@ object DeletedKubernetesClusterChecker {
           _ <- if (isDryRun) F.unit
           else
             clusterOpt match {
-              case None    => dbReader.updateK8sClusterStatus(cluster.id)
+              case None    => dbReader.markK8sClusterDeleted(cluster.id) >> dbReader.unlinkPDFromK8sCluster(cluster.id)
```
Qi77Qi, line 36: maybe we can just null `diskid` in `markK8sClusterDeleted` query?

core/src/test/scala/com/broadinstitute/dsp/DBTestHelper.scala
ehanna4: Sorry about that, I removed `stopAfterCreation` and it's corresponding value which were causing the unit tests to fail since this field had been removed from the leo API (in a previous ticket not related to this)

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedKubernetesClusterChecker.scala
ehanna4: Hmm I can revisit and try that if you think that'd be better, my thought was to keep the queries separate to be the most "single purpose" but I am not very strongly opinionated on it.

Qi77Qi: I agree in general, but I think in this case, whenever we mark a cluster deleted, we probably do want to unlink disk as well

ehanna4: Okay I have combined the queries to unlink and delete in the same query for each K8s cluster and runtimes

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
```diff
@@ -55,10 +55,13 @@ object DbReader {
            update PERSISTENT_DISK set status = "Deleted", destroyedDate = now() where id = $id
            """.update
 
-  def updateK8sClusterStatusQuery(id: Int) =
+  def markK8sClusterDeletedQuery(id: Int) =
     sql"""
-           update KUBERNETES_CLUSTER set status = "DELETED", destroyedDate = now() where id = $id
-           """.update
+          UPDATE NODEPOOL
```
Qi77Qi, line 60: oh, didn't realize it needs to join so many tables..I probbly would've been fine with what you had before :) ...

Qi77Qi (approved):
thanks for addressing the comments!

PR #48, opened 2021-04-07
IA-2589 support multi zone and region

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-2589

* Add multi zone and region support. Since disk zone is not persisted yet, so there're still a few default zone references left. Those should be addressed after https://broadworkbench.atlassian.net/browse/IA-2640
* Added `cronJobs.dbTest` for running all DB access unit tests. Altho, tests don't seem to pass when running at the same time due to some errors like. All db unit tests are passing when running separately. Will log a separate ticket to figure out how these DB tests can be run together.
```
Cannot add or update a child row: a foreign key constraint fails (`leotestdb`.`NODEPOOL`, CONSTRAINT `FK_NODEPOOL_CLUSTER_ID` FOREIGN KEY (`clusterId`) REFERENCES `KUBERNETES_CLUSTER` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION)
```

### DIFF
--- a/.github/workflows/unitTests.yml
+++ b/.github/workflows/unitTests.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Run tests
-        run: sbt coverage test coverageReport
+        run: sbt coverage "testOnly -- -l cronJobs.dbTest" coverageReport
       - name: Check output
         run: |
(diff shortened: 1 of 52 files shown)


rtitle (approved):
LGTM.. thanks for updating this.

PR #61, opened 2021-04-08
Update Resource Validator to handle the 2nd Galaxy disk (IA-2741)

ehanna4:
Update ResourceValidator to delete metadata disks from Google if cleaning up the corresponding data disk. Relies on this PR for Leo (https://github.com/DataBiosphere/leonardo/pull/2069) to update the naming convention for Galaxy metadata disks to be discoverable using the PERSISTENT_DISK table in the Leo DB.

Tested successfully on dev
Made a Galaxy app then detached the disks in Google and set them to "deleted" status in the DB. Ran the cronjob and both disks were deleted from Google
Before:
Screen Shot 2021-05-19 at 4 01 08 PM
After:
Screen Shot 2021-05-19 at 4 02 13 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,11 +50,22 @@ e.g. `sbt "resourceValidator/run --dryRun --all"`
 
 ## Contributing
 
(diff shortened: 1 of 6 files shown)


Qi77Qi (approved):
lgtm 👍

PR #63, opened 2021-04-08
[IA-2644] remove reference to batch nodepools

jdcanas:


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -160,7 +158,7 @@ object DbReader {
               NOT EXISTS (
                 SELECT *
                 FROM NODEPOOL np
-                LEFT JOIN APP a ON np.id = a.nodepoolId
+                RIGHT JOIN APP a ON np.id = a.nodepoolId
```
jdcanas, line 161: This changes it to remove clusters with no nodepools and no apps.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -160,7 +160,7 @@ object DbReader {
...
                 WHERE
                   kc.id = np.clusterId AND np.isDefault = 0 AND
                   (
```
rtitle, line 166: Can probably remove `OR (a.id IS NULL)` at the end because that's impossible with a right join, correct?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -107,7 +107,6 @@ object DbReader {
   // Note that we explicitly check nodepools with 5/11 of statuses that exist.
   // The statuses we exclude are PROVISIONING, STOPPING, DELETED, PRECREATING, PREDELETING, and PREDELETING
   //    - We exclude all -ING statuses because they are transitionary, and the purpose of this is not to handle timeouts
-  //    - Unclaimed we exclude because we never want to clean up batch created nodepools
```
kyuksel, line 110: Wasn't there something in `applessNodepoolQuery` that disallowed batch-created nodepools to be selected by the query?

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/DbReaderGetKubernetesClustersToDeleteSpec.scala
```diff
@@ -198,8 +198,7 @@ final class DbReaderGetKubernetesClustersToDeleteSpec extends AnyFlatSpec with C
     }
   }
 
-  // The scenario below is to mimic a cluster with batch-pre-created nodepools, which we don't want to auto-delete
-  it should "NOT detect for removal: Kubernetes cluster with nodepools but no apps" taggedAs DbTest in {
+  it should "detect for removal: Kubernetes cluster with nodepools but no apps" taggedAs DbTest in {
```
kyuksel, line 201: My expectations:
1. The **older** version of this test should fail **with** your change in `kubernetesClustersToDeleteQuery`.
2. The **newer** version of this test should fail **without** your change in `kubernetesClustersToDeleteQuery`.

Have you observed these?

kyuksel (commented):
From past experience, I recall it was really easy to make mistakes with these very complicated queries so I'd recommend heavily relying on unit testing and hence my comments inline.

resource-validator/src/test/scala/com/broadinstitute/dsp/resourceValidator/DbReaderGetKubernetesClustersToDeleteSpec.scala
jdcanas: Yes, I played around with switching it back and forth

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
jdcanas: Yeah... its definitely confusing. I changed it back and forth and ran tests repeatedly

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
jdcanas: I'm not super sure... currently, this query matches cluster with nodepools and no apps. Not familiar enough with SQL to know if `a.id` would be null in that scenario.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
jdcanas: I see a test for this a bit unecessary since the status in question (NodepoolStatus.Unclaimed) was removed from leo entirely

jdcanas: Yeah, a change was needed for applessNodepoolQuery https://github.com/broadinstitute/leonardo-cron-jobs/pull/63/files#diff-7bd14f4fe178f136127d110beb4c945537788ea7bc7f4c066245c660219a9427R163

kyuksel: The change you linked to is for `kubernetesClustersToDeleteQuery`, not `applessNodepoolQuery`. 

Do you remember why the comment

> Unclaimed we exclude because we never want to clean up batch created nodepools

was put there at the first place? If it was there by mistake, we're good but if not, it'd make me wonder if the query needs updating to not do any exclusion anymore.

jdcanas: Ah, sorry. Yeah, so the query explicitly selects the statuses we want to include. 'Unclaimed' is not there for the reason described, which is no longer relevant, so only a comment update was necessary. The query remains correct still.

jdcanas:
> From past experience, I recall it was really easy to make mistakes with these very complicated queries so I'd recommend heavily relying on unit testing and hence my comments inline.

Yeah... I definitely had to rely on unit tests.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
kyuksel: Ah now I understand - thank you.

kyuksel (approved):
👍🏼

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/KubernetesClusterRemover.scala
```diff
@@ -34,7 +34,7 @@ object KubernetesClusterRemover {
       override def resourceToScan: fs2.Stream[F, KubernetesClusterToRemove] = dbReader.getKubernetesClustersToDelete
 
       // TODO: This check is to be moved to a new project (a.k.a. 'janitor)
-      // https://broadworkbench.atlassian.net/wiki/spaces/IA/pages/807436289/2020-09-17+Leonardo+Async+Processes
+      // https:x//broadworkbench.atlassian.net/wiki/spaces/IA/pages/807436289/2020-09-17+Leonardo+Async+Processes
```
Qi77Qi, line 37: extra `x`?

PR #164, opened 2021-12-20
IA-3005  update dataproc cluster to fully stop clusters

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3005

1. Update resource validator to use the new dataproc cluster client so that it'll fully stop clusters.. See `StoppedRuntimeChecker`
2. Update zombie monitor in a way that if it detects an active runtime (active in leo DB) is `Stopped` in google, it'll update DB to reflect it...See `ActiveRuntimeChecker`. 
3. Fix a bug in `markK8sClusterDeletedQuery` query. We can't `INNER JOIN` nodepool table and App table because not all nodepools have an App. We can't `INNER JOIN` kubernetes cluster table and nodepool table either because not all cluster id has a corresponding nodepool.

Majority of the diff is due to cats-effect 3 migration, and some leftover refactoring needed due to azure refactoring..

Tested all runners locally against dev

### DIFF
--- a/core/src/main/scala/com/broadinstitute/dsp/CheckRunner.scala
+++ b/core/src/main/scala/com/broadinstitute/dsp/CheckRunner.scala
@@ -1,18 +1,16 @@
 package com.broadinstitute.dsp
 
-import java.nio.charset.Charset
-import java.time.Instant
-import java.util.concurrent.TimeUnit
-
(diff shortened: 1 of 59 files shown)


gpcarr (approved):
LGTM :shipit:

PR #174, opened 2022-01-06
IA-3005 fix shutdown issue

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3005

In current code, because `goog-compute-???` threads aren't daemon threads, they'll prevent JVM from shutting down properly. It shutsdown just fine via `sbt` for some reason, but issue is exposed when deployed running in a container.

See error
```
Non-daemon threads currently preventing JVM termination: - 299: Thread[DestroyJavaVM,5,main]
```

Depends on https://github.com/broadinstitute/workbench-libs/pull/864


kyuksel (approved):
LGTM 👍🏼

PR #175, opened 2022-01-11
IA-3005 bump wb-libs

Qi77Qi:
Similar to https://github.com/broadinstitute/leonardo-cron-jobs/pull/174.

Relevant wb-libs PR https://github.com/broadinstitute/workbench-libs/pull/872

The previous PR fixed zombie monitor. But the resourceValidator and janitor uses pubsub, which also uses non daemon threads.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -1,7 +1,7 @@
 import sbt._
 
 object Dependencies {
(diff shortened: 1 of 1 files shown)


xuscrbw (approved):
lgtm

PR #192, opened 2022-03-27
[no ticket] not try to delete kubernetes cluster if billing is disabled

Qi77Qi:
In Dev, we keep getting messages about delete kubernetes clusters. Those clusters live in billing projects that are already disabled, so when leo tries to disable, things will error out. Next time cron job runs, these clusters will get picked up again, and leo will fail to delete again.

This changes the cron job to ignore clusters if its project doesn't have billing enabled


xuscrbw (approved):
lgtm

Qi77Qi:
thanks @xuscrbw !

PR #194, opened 2022-03-29
[IA-3281] Fixing Resource Validator

xuscrbw:
Noticing these errors in the last few runs: 2022-03-28 04:48:44.937 EDTdoobie.util.invariant$InvalidValue: AZURE_VM: String invalid for com.broadinstitute.dsp.CloudService because: invalid cloudService value AZURE_VM

Added a TODO for once we support Azure clean up

### DIFF
--- a/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
+++ b/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
@@ -6,7 +6,8 @@ import org.scalacheck.{Arbitrary, Gen}
 import org.broadinstitute.dsde.workbench.google2.Generators._
 
 object Generators {
-  val genCloudService: Gen[CloudService] = Gen.oneOf(CloudService.Gce, CloudService.Dataproc, CloudService.AzureVM)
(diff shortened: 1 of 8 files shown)


core/src/test/scala/com/broadinstitute/dsp/Generators.scala
```diff
@@ -30,12 +33,7 @@ object Generators {
     project <- genGoogleProject
     diskName <- genDiskName
     zone <- genZoneName
-  } yield Disk(id,
-               CloudContext.Gcp(project),
-               diskName,
-               zone,
-               formattedBy = Some("GCE")
-  ) // TODO: update generator once we support Azure
+  } yield Disk(id, CloudContext.Gcp(project), diskName, zone, formattedBy = Some("GCE"))
```
Qi77Qi, line 36: think the todo is still valid

Qi77Qi (approved):
thanks @xuscrbw ..we should run this once it's deployed

Qi77Qi (approved):

PR #204, opened 2022-06-17
IA-3480 delete "Deleting" disk as well

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3480

Disks that're in "Deleting" status means at some point there was user request to delete these disks, but somehow they're stuck in deleting. So we're discovering these disks and deleting them in GCP.

Once these disks are actually deleted in GCP, they'll be picked up by zombie monitor and leo DB will reflect the right status from then on.

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
@@ -22,17 +22,20 @@ trait DbReader[F[_]] {
 object DbReader {
   implicit def apply[F[_]](implicit ev: DbReader[F]): DbReader[F] = ev
 
-  val deletedDisksQuery =
+  val shouldBedeletedDisksQuery =
     sql"""
            SELECT pd1.id, pd1.cloudContext, pd1.cloudProvider, pd1.name, pd1.zone, pd1.formattedBy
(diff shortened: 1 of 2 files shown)


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -22,17 +22,20 @@ trait DbReader[F[_]] {
...
+  val shouldBedeletedDisksQuery =
     sql"""
            SELECT pd1.id, pd1.cloudContext, pd1.cloudProvider, pd1.name, pd1.zone, pd1.formattedBy
            FROM PERSISTENT_DISK AS pd1
-           WHERE pd1.status="Deleted" AND
-           pd1.destroyedDate > now() - INTERVAL 30 DAY AND
+           WHERE
+           (
+             (pd1.status="Deleted" AND pd1.destroyedDate > now() - INTERVAL 30 DAY) OR 
+             (pd1.status="Deleting" AND pd1.`dateAccessed` < now() - INTERVAL 30 MINUTE)
```
Qi77Qi, line 32: if a disk is stuck in "Deleting" status for more than 30 min, we consider this an anomaly and resource validator will intervene and delete it

xuscrbw (approved):
LGTM

PR #205, opened 2022-06-28
IA-3289 address azure runtime as well in resource-validator

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3289

This PR mostly brings in azure lib in https://github.com/broadinstitute/workbench-libs/pull/1005, and addressed a few gaps in checkers for Azure runtimes.

The only thing left is for checking "stopping" status since we don't currently support start/stop azure runtimes yet.

Tested resourceValidator locally
`sbt:resource-validator> run --dryRun --all`

Will keep an eye on things once it's merged to make sure all checks are working as expected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ target/
 application.conf
 sqlproxy.env
 .bsp
+application.conf.qi
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
@@ -44,11 +44,15 @@ object StoppedRuntimeChecker {
           runtimeOpt <- deps.computeService
             .getInstance(runtime.googleProject, runtime.zone, InstanceName(runtime.runtimeName))
           runningRuntimeOpt <- runtimeOpt.flatTraverse { rt =>
-            if (rt.getStatus.toUpperCase == Instance.Status.RUNNING.name().toUpperCase)
+            val expectedStatus =
+              Set(Instance.Status.STOPPED.name().toUpperCase, Instance.Status.TERMINATED.name().toUpperCase())
+            if (!expectedStatus.contains(rt.getStatus.toUpperCase))
               if (isDryRun)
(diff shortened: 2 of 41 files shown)


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
```diff
@@ -43,7 +44,7 @@ object StoppedRuntimeChecker {
           runtimeOpt <- deps.computeService
             .getInstance(runtime.googleProject, runtime.zone, InstanceName(runtime.runtimeName))
           runningRuntimeOpt <- runtimeOpt.flatTraverse { rt =>
-            if (rt.getStatus == Instance.Status.RUNNING)
+            if (rt.getStatus.toUpperCase == Instance.Status.RUNNING.name().toUpperCase)
```
gpcarr, line 47: How do we handle other statuses for GCE or Dataproc? I know Galaxy is the only resource we do all uppercase for in terms status - do we do `.toUpperCase` everywhere to be consistent?

.gitignore
```diff
@@ -4,4 +4,5 @@ target/
 .idea/
 application.conf
 sqlproxy.env
-.bsp
\ No newline at end of file
+.bsp
+application.conf.qi
```
gpcarr, line 8: do you have a local `application.conf.qi` that you use to test these changes? is that config documented anywhere?

Qi77Qi: I don't actually have it anymore...but good to have it in .gitignore just so I don't have to worry about a copy of my own config file

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
Qi77Qi: the upper case here is just for this particular comparison...we could use `toLowerCase` here too..the point is to make sure casing is consistent for both left and right side

> How do we handle other statuses for GCE or Dataproc? 
This check is to make sure all `Stopped` should be indeed `Stopped` in GCP. I guess I can make the check to error if the VM status is not `Stopped` or `Terminated` to make it a bit more inclusive

Qi77Qi:
@gpcarr lmk if you have more comments

gpcarr (approved):
:shipit:

PR #210, opened 2022-08-12
IA-3631 Fix assembly

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3631

Resource validator is currently broken becuz I think `sbt assembly` removed some of the classes.


rtitle (approved):
Looks fine.. the other way to fix these types of issues is to add sbt exclusions. E.g. not sure we need `bouncycastle`.

PR #238, opened 2022-12-21
Combine all dependencies updates since 10-08-2022

LizBaldo:
Once this passes, I will close any of the previous PRs.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.7.1
+sbt.version=1.7.3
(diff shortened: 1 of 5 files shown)


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.7.1
+sbt.version=1.7.3
```
Qi77Qi, line 1: think sbt 1.8.0 has come out :)

Qi77Qi (approved):
looks great...thanks!

PR #256, opened 2023-02-15
[IA-3623] Update cron jobs to support AKS apps

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-3623

While working on the [Leo Apps Dashboard](https://grafana.dsp-devops.broadinstitute.org/d/mmgNHw04z/leonardo-apps?from=now-1h&to=now&var-cluster=terra-dev&var-namespace=terra-dev&var-uiClient=All&var-cloud=All&var-mrg=All&orgId=1) it became apparent that there are many Azure apps in dev in active statuses (`RUNNING`, `PROVISIONING`, etc) whose MRGs have been deleted. 

So I decided to pick up [IA-3623](https://broadworkbench.atlassian.net/browse/IA-3623) and implement the zombie-monitor cronjobs for AKS apps. The PR expanded to some additional refactoring but hopefully it's not too bad.

Brief description of changes (more context inline):
1. Slight model refactoring, remove `K8sClusterToScan` and `NodepoolToScan` classes in favor of `KubernetesCluster` and `Nodepool` which support both clouds.
2. Enabled Azure generators for unit tests, which caused some test refactoring but had to be done, I think.
3. The _actual_ change is in the cluster/nodepool checkers in `zombie-monitor`.
   - `resource-validator` and `janitor` still have Azure k8s checks disabled, I don't think it really makes sense to enable them given our AKS implementation.

I tested via unit tests; running `sbt "zombieMonitor/run --dryRun --all"` a few times; and even running a non-dryRun in dev 1 time. I think it works by inspecting the DB and checking grafana.

[IA-3623]: https://broadworkbench.atlassian.net/browse/IA-3623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/test/scala/com/broadinstitute/dsp/DbTestHelper.scala
+++ b/core/src/test/scala/com/broadinstitute/dsp/DbTestHelper.scala
@@ -5,7 +5,6 @@ import doobie.Put
 import doobie.hikari.HikariTransactor
 import doobie.implicits._
 import doobie.util.transactor.Transactor
-import org.broadinstitute.dsde.workbench.google2.GKEModels.KubernetesClusterId
 import org.broadinstitute.dsde.workbench.google2.KubernetesSerializableName.NamespaceName
 import org.broadinstitute.dsde.workbench.google2.{RegionName, ZoneName}
 import org.scalatest.Tag
@@ -42,12 +41,12 @@ object DbTestHelper {
   def insertDisk(disk: Disk, status: String = "Ready")(implicit xa: HikariTransactor[IO]): IO[Long] =
     insertDiskQuery(disk, status: String).withUniqueGeneratedKeys[Long]("id").transact(xa)
 
-  def insertK8sCluster(clusterId: KubernetesClusterId, status: String = "RUNNING")(implicit
+  def insertK8sCluster(cluster: KubernetesCluster, status: String = "RUNNING")(implicit
     xa: HikariTransactor[IO]
   ): IO[Long] =
     sql"""INSERT INTO KUBERNETES_CLUSTER
          (cloudContext, clusterName, location, status, creator, createdDate, destroyedDate, dateAccessed, loadBalancerIp, networkName, subNetworkName, subNetworkIpRange, region, apiServerIp, ingressChart)
-         VALUES (${clusterId.project}, ${clusterId.clusterName}, ${clusterId.location}, ${status}, "fake@broadinstitute.org", now(), now(), now(), "0.0.0.1", "network", "subnetwork", "0.0.0.1/20", ${regionName}, "35.202.56.6", "stable/nginx-ingress-1.41.3")
+         VALUES (${cluster.cloudContext.asString}, ${cluster.clusterName}, ${cluster.location}, ${status}, "fake@broadinstitute.org", now(), now(), now(), "0.0.0.1", "network", "subnetwork", "0.0.0.1/20", ${regionName}, "35.202.56.6", "stable/nginx-ingress-1.41.3")
          """.update.withUniqueGeneratedKeys[Long]("id").transact(xa)
 
   def insertNodepool(clusterId: Long, nodepoolName: String, isDefault: Boolean, status: String = "RUNNING")(implicit
--- a/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
+++ b/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
@@ -40,7 +40,7 @@ object Generators {
                diskName,
                zone,
                formattedBy = Some("GCE")
-  ) // TODO: update generator once we support Azure disks
+  ) // TODO: update generator once we support Azure
 
   val genInitBucket: Gen[InitBucketToRemove] = for {
     project <- genGoogleProject
--- a/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
+++ b/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
@@ -7,8 +7,7 @@ import org.broadinstitute.dsde.workbench.google2.Generators._
 import org.scalacheck.{Arbitrary, Gen}
 
 object Generators {
-  // TODO IA-3289 When we implement Azure, make sure to add CloudService.AzureVM as an option in the line below so tests use it
-  val genCloudService: Gen[CloudService] = Gen.oneOf(CloudService.Gce, CloudService.Dataproc)
+  val genCloudService: Gen[CloudService] = Gen.oneOf(CloudService.Gce, CloudService.Dataproc, CloudService.AzureVM)
   def genRuntime(possibleStatuses: Option[NonEmptyList[String]]): Gen[Runtime] = for {
     id <- Gen.chooseNum(0, 100)
     cloudService <- genCloudService
--- a/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
+++ b/core/src/test/scala/com/broadinstitute/dsp/Generators.scala
@@ -1,10 +1,10 @@
 package com.broadinstitute.dsp
 
 import cats.data.NonEmptyList
-import org.broadinstitute.dsde.workbench.google2.GKEModels.{KubernetesClusterId, KubernetesClusterName}
-import org.scalacheck.{Arbitrary, Gen}
-import org.broadinstitute.dsde.workbench.google2.Generators._
 import org.broadinstitute.dsde.workbench.azure.Generators.genAzureCloudContext
+import org.broadinstitute.dsde.workbench.google2.GKEModels.KubernetesClusterName
+import org.broadinstitute.dsde.workbench.google2.Generators._
+import org.scalacheck.{Arbitrary, Gen}
 
 object Generators {
   // TODO IA-3289 When we implement Azure, make sure to add CloudService.AzureVM as an option in the line below so tests use it
@@ -40,7 +40,7 @@ object Generators {
                diskName,
                zone,
                formattedBy = Some("GCE")
-  ) // TODO: update generator once we support Azure
+  ) // TODO: update generator once we support Azure disks
 
   val genInitBucket: Gen[InitBucketToRemove] = for {
     project <- genGoogleProject
@@ -48,15 +48,16 @@ object Generators {
   } yield InitBucketToRemove(project, Some(InitBucketName(bucketName.value)))
 
   val genKubernetesCluster: Gen[KubernetesCluster] = for {
+    id <- Gen.chooseNum(0, 100)
     name <- Gen.uuid.map(x => KubernetesClusterName(x.toString))
     cloudService <- genCloudService
     project <- genGoogleProject
     azureCloudContext <- genAzureCloudContext
     location <- genLocation
   } yield cloudService match {
-    case CloudService.Gce      => KubernetesCluster(name, CloudContext.Gcp(project), location)
-    case CloudService.Dataproc => KubernetesCluster(name, CloudContext.Gcp(project), location)
-    case CloudService.AzureVM  => KubernetesCluster(name, CloudContext.Azure(azureCloudContext), location)
+    case CloudService.Gce      => KubernetesCluster(id, name, CloudContext.Gcp(project), location)
+    case CloudService.Dataproc => KubernetesCluster(id, name, CloudContext.Gcp(project), location)
+    case CloudService.AzureVM  => KubernetesCluster(id, name, CloudContext.Azure(azureCloudContext), location)
   }
 
   val genNodepool: Gen[Nodepool] = for {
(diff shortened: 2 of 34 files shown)


core/src/test/scala/com/broadinstitute/dsp/Generators.scala
```diff
@@ -1,14 +1,13 @@
...
-import org.broadinstitute.dsde.workbench.google2.Generators._
 import org.broadinstitute.dsde.workbench.azure.Generators.genAzureCloudContext
+import org.broadinstitute.dsde.workbench.google2.GKEModels.KubernetesClusterName
+import org.broadinstitute.dsde.workbench.google2.Generators._
+import org.scalacheck.{Arbitrary, Gen}
 
 object Generators {
-  // TODO IA-3289 When we implement Azure, make sure to add CloudService.AzureVM as an option in the line below so tests use it
-  val genCloudService: Gen[CloudService] = Gen.oneOf(CloudService.Gce, CloudService.Dataproc)
+  val genCloudService: Gen[CloudService] = Gen.oneOf(CloudService.Gce, CloudService.Dataproc, CloudService.AzureVM)
```
rtitle, line 10: I think the unit tests weren't actually testing Azure before -- now they are. :)

janitor/src/main/scala/com/broadinstitute/dsp/janitor/NodepoolRemover.scala
```diff
@@ -24,17 +24,15 @@ object NodepoolRemover {
...
-                case CloudContext.Gcp(value) =>
-                  val msg = DeleteNodepoolMeesage(n.nodepoolId, value, Some(ctx))
-                  deps.publisher.publishOne(msg)
-                case CloudContext.Azure(_) =>
-                  logger.warn("Azure is not supported yet") // TODO: IA-3623
-              }
-
-            } else F.unit
-        } yield Some(n)
+          res <-
```
rtitle, line 27: Changed this logic (kind of inverted the `isDryRun` and `cloudContext` check) to be consistent with the others. Now we only return `Some` for the GCP case; and only publish if `isDryRun` is true.

project/Dependencies.scala
```diff
@@ -1,8 +1,8 @@
 import sbt._
 
 object Dependencies {
-  val workbenchLibsHash = "0545bbb3-SNAP"
-  val workbenchGoogle2Version = s"0.24-${workbenchLibsHash}"
+  val workbenchLibsHash = "87b3b166-SNAP"
```
rtitle, line 4: https://github.com/broadinstitute/workbench-libs/pull/1249

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/StoppedRuntimeChecker.scala
```diff
@@ -34,7 +34,7 @@ object StoppedRuntimeChecker {
             checkGceRuntime(x, isDryRun)
           case _: Runtime.AzureVM =>
             // TODO: IA-3289 Implement check Azure VM
-            logger.info(s"Azure VM is not supported yet").as(None)
+            logger.info(s"StoppedRuntimeChecker not supported for Azure VMs").as(None)
```
rtitle, line 37: This one should be implemented once we fix pause/resume.

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletedOrErroredKubernetesClusterChecker.scala
```diff
@@ -47,7 +47,8 @@ object DeletedOrErroredKubernetesClusterChecker {
                 }
               }
             } yield clusterOpt.fold(none[KubernetesCluster])(_ => Some(cluster))
-          case CloudContext.Azure(_) => logger.warn("Azure is not supported yet").as(none)
+          case CloudContext.Azure(_) =>
+            logger.warn("Resource validator not supported for Azure k8s clusters").as(none)
```
rtitle, line 51: As mentioned above I don't think `resourceValidator` ever really makes sense for our AKS implementation. The reason is that Leo doesn't own the AKS/nodepool provisioning -- it's all part of the landing zone.

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedKubernetesClusterChecker.scala
```diff
@@ -16,29 +18,57 @@ object DeletedKubernetesClusterChecker {
...
       override def configs = CheckRunnerConfigs(s"deleted-kubernetes", false)
 
       override def dependencies: CheckRunnerDeps[F] = deps.checkRunnerDeps
 
-      def checkResource(cluster: K8sClusterToScan, isDryRun: Boolean)(implicit
+      def checkResource(cluster: KubernetesCluster, isDryRun: Boolean)(implicit
         ev: Ask[F, TraceId]
-      ): F[Option[K8sClusterToScan]] =
+      ): F[Option[KubernetesCluster]] = {
+        val isZombie = cluster.cloudContext match {
```
rtitle, line 34: Actual functional change in this file: it's now checking both GCP and Azure k8s clusters for zombie-ness.

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedOrErroredNodepoolChecker.scala
```diff
@@ -19,35 +21,64 @@ object DeletedOrErroredNodepoolChecker {
...
       override def configs = CheckRunnerConfigs(s"deleted-or-errored-nodepools", false)
 
       override def dependencies: CheckRunnerDeps[F] = deps.checkRunnerDeps
 
-      def checkResource(nodepoolToScan: NodepoolToScan, isDryRun: Boolean)(implicit
+      def checkResource(nodepool: Nodepool, isDryRun: Boolean)(implicit
         ev: Ask[F, TraceId]
-      ): F[Option[NodepoolToScan]] =
+      ): F[Option[Nodepool]] = {
+        val isZombie = nodepool.cloudContext match {
```
rtitle, line 38: Functional change -- checking both GCP and Azure nodepool for zombie.

core/src/test/scala/com/broadinstitute/dsp/DbTestHelper.scala
```diff
@@ -42,12 +41,12 @@ object DbTestHelper {
   def insertDisk(disk: Disk, status: String = "Ready")(implicit xa: HikariTransactor[IO]): IO[Long] =
     insertDiskQuery(disk, status: String).withUniqueGeneratedKeys[Long]("id").transact(xa)
 
-  def insertK8sCluster(clusterId: KubernetesClusterId, status: String = "RUNNING")(implicit
+  def insertK8sCluster(cluster: KubernetesCluster, status: String = "RUNNING")(implicit
```
LizBaldo, line 44: Damn, this change from `KubernetesClusterId` to `KubernetesCluster` is super powerful. Is there an existing documentation you could point me to so I can read up on that?

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
```diff
@@ -128,8 +127,8 @@ object DbReader {
     override def markK8sClusterDeleted(id: Long): F[Unit] =
       markK8sClusterDeletedQuery(id.toInt).run.transact(xa).void
 
-    override def getk8sNodepoolsToDeleteCandidate: Stream[F, NodepoolToScan] =
-      activeNodepoolsQuery.stream.unNone.transact(xa)
```
LizBaldo, line 132: Why remove this?

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedOrErroredNodepoolChecker.scala
```diff
@@ -19,35 +21,64 @@ object DeletedOrErroredNodepoolChecker {
...
+      ): F[Boolean] =
+        for {
+          nodepoolOpt <- deps.gkeService.getNodepool(gkeNodepoolId)
+          deleted <-
+            nodepoolOpt match {
+              case None =>
+                logger.info(s"Going to mark GKE nodepool ${id} and apps as DELETED") >>
+                  (if (!isDryRun) dbReader.markNodepoolAndAppDeleted(id) else F.unit) >>
+                  F.pure(true)
+              case Some(nodepool) if nodepool.getStatus == com.google.container.v1.NodePool.Status.ERROR =>
```
LizBaldo, line 61: I know this is pre-existing logic, but why do we change the status to error here?

LizBaldo: Especially since we do not do it for Azure, should we be consistent across cloud platforms?

LizBaldo (approved):
That is a wonderful refactoring thank you so much for doing this. It makes perfect sense to kickstart with the zombie-monitor in order to have the leo grafana views actually reflect the resources that we are using.

I left a few comments/questions that should not really block you from merging. We might be getting the StuckAppReporter in as a bonus as well.

I would like to repurpose this ticket https://broadworkbench.atlassian.net/browse/IA-4022 to investigate enabling azure for the resource-validator and the janitor jobs to reduce the amount of resources leaking. What do you think?

janitor/src/main/scala/com/broadinstitute/dsp/janitor/KubernetesClusterRemover.scala
```diff
@@ -50,7 +50,7 @@ object KubernetesClusterRemover {
                   } else F.unit
                 res = if (isBillingEnabled) Some(c) else None
               } yield res
-            case CloudContext.Azure(_) => logger.warn("Azure k8s clusterRemover is not supported yet").as(none)
```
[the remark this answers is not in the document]
rtitle: Can chat more in person, but my thinking is basically as follows. On GCP Leo controls the provisioning of the GKE cluster and nodepool. When all apps in a cluster have been deleted, the Janitor cronjob detects this and "garbage collects" the cluster/nodepool, by sending Leo a pubsub message to delete it.

On Azure Leo does not control the provisioning of the AKS cluster + nodepool. The cluster is provisioned by the Landing Zone, and is always running. So Leo is just concerned with apps, and there is no need to "garbage collect" the cluster/nodepool.

core/src/test/scala/com/broadinstitute/dsp/DbTestHelper.scala
rtitle: `KubernetesClusterId` comes from here in workbench-libs/google2: https://github.com/broadinstitute/workbench-libs/blob/develop/google2/src/main/scala/org/broadinstitute/dsde/workbench/google2/kubernetesModels.scala#L95. It is really a GKE-specific thing.

`KubernetesCluster` is a model in the cronjobs repo, and supports both clouds.

The naming is so similar, and confused me for a bit. 😅 No real documentation except staring at code.

….And maybe worth chatting about later, I feel like the Leo data model for k8s cluster/nodepool/app in general. The code is kind of confusing and the data model is hard to work with. But not high priority, maybe future tech debt.

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DbReader.scala
rtitle: oh the `unNone`? 

I think the query to list `NodepoolToScan` records returned an option because it explicitly didn't handle Azure. So the stream here removed all the Nones. Now that we're using `Nodepool` we handle both clouds. 😀

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedOrErroredNodepoolChecker.scala
rtitle: Yeah I guess if the GCP nodepool is Error, we set the Leo status to Error. I honestly don't remember why…. maybe we wanted to surface the cloud error to users instead of having it disappear?

Azure is kind of "dumber" in this regard -- Leo does not manage clusters/nodepools so here we're just testing the cloud cluster for existence, not whether it errored. I think this is reasonable to start with, but maybe could get more sophisticated if we see Azure nodepools start erroring.

PR #289, opened 2023-08-21
[IA-4459] Clean up runtimes stuck in deleting status for over an hour

LizBaldo:
Jira ticket: https://broadworkbench.atlassian.net/browse/IA-4459

Note that I still need to test, but I need to figure out how to run this on my BEE first :)

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
@@ -11,6 +11,7 @@ import org.broadinstitute.dsde.workbench.model.google.{GcsBucketName, GoogleProj
 trait DbReader[F[_]] {
   def getDeletedDisks: Stream[F, Disk]
   def getDeletedRuntimes: Stream[F, Runtime]
+  def getDeletingRuntimes: Stream[F, Runtime]
   def getErroredRuntimes: Stream[F, Runtime]
   def getStoppedRuntimes: Stream[F, Runtime]
   def getInitBucketsToDelete: Stream[F, InitBucketToRemove]
@@ -50,6 +51,23 @@ object DbReader {
           WHERE
             c1.status = "Deleted" AND
             c1.destroyedDate > now() - INTERVAL 30 DAY AND
+            NOT EXISTS (
+              SELECT *
+              FROM CLUSTER AS c2
+              WHERE
+                c2.cloudContext = c1.cloudContext AND
+                c2.runtimeName = c1.runtimeName AND
+                (c2.status = "Stopped" OR c2.status = "Running")
+          )"""
+      .query[Runtime]
+
+  val deletingRuntimeQuery =
+    sql"""SELECT DISTINCT c1.id, cloudContext, c1.cloudProvider, runtimeName, rt.cloudService, c1.status, rt.zone, rt.region
+          FROM CLUSTER AS c1
+          INNER JOIN RUNTIME_CONFIG AS rt ON c1.runtimeConfigId = rt.id
+          WHERE
+            c1.status = "Deleting" AND
+            c1.destroyedDate > now() - INTERVAL 60 MIN AND
             NOT EXISTS (
               SELECT *
               FROM CLUSTER AS c2
@@ -128,6 +146,9 @@ object DbReader {
      */
     override def getDeletedRuntimes: Stream[F, Runtime] =
       deletedRuntimeQuery.stream.transact(xa)
+
+    override def getDeletingRuntimes: Stream[F, Runtime] =
+      deletingRuntimeQuery.stream.transact(xa)
 
     override def getErroredRuntimes: Stream[F, Runtime] =
       erroredRuntimeQuery.stream.transact(xa)
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletingRuntimeChecker.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletingRuntimeChecker.scala
@@ -1,0 +1,104 @@
+package com.broadinstitute.dsp
+package resourceValidator
+
+import cats.effect.Concurrent
+import cats.mtl.Ask
+import cats.syntax.all._
+import org.broadinstitute.dsde.workbench.google2.DataprocClusterName
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.util2.InstanceName
+import org.typelevel.log4cats.Logger
+
+// Implements CheckRunner[F[_], A]
+object DeletingRuntimeChecker {
+  def impl[F[_]](
+    dbReader: DbReader[F],
+    deps: RuntimeCheckerDeps[F]
+  )(implicit F: Concurrent[F], logger: Logger[F], ev: Ask[F, TraceId]): CheckRunner[F, Runtime] =
+    new CheckRunner[F, Runtime] {
+      override def appName: String = resourceValidator.appName
+
+      override def configs = CheckRunnerConfigs(s"deleting-runtime", shouldAlert = true)
+
+      override def dependencies: CheckRunnerDeps[F] = deps.checkRunnerDeps
+
+      override def resourceToScan: fs2.Stream[F, Runtime] = dbReader.getDeletingRuntimes
+
+      override def checkResource(runtime: Runtime, isDryRun: Boolean)(implicit
+        ev: Ask[F, TraceId]
+      ): F[Option[Runtime]] =
+        runtime match {
+          case x: Runtime.Dataproc =>
+            checkDataprocCluster(x, isDryRun)
+          case x: Runtime.Gce =>
+            checkGceRuntime(x, isDryRun)
+          case x: Runtime.AzureVM =>
+            checkAzureRuntime(x, isDryRun)
+        }
+
+      private def checkDataprocCluster(runtime: Runtime.Dataproc, isDryRun: Boolean)(implicit
+        ev: Ask[F, TraceId]
+      ): F[Option[Runtime]] =
+        for {
+          clusterOpt <- deps.dataprocService
+            .getCluster(runtime.googleProject, runtime.region, DataprocClusterName(runtime.runtimeName))
+          r <- clusterOpt.flatTraverse { _ =>
+            for {
+              isBillingEnabled <- deps.billingService.isBillingEnabled(runtime.googleProject)
+              r <-
+                if (isDryRun)
+                  logger
+                    .warn(
+                      s"$runtime still exists in Google. It needs to be deleted. isBillingEnabled: $isBillingEnabled. Project: ${runtime.googleProject}"
+                    )
+                    .as(runtime.some)
+                else
+                  isBillingEnabled match {
+                    case true =>
+                      logger.warn(
+                        s"$runtime still exists in Google and billing is enabled. Going to delete it."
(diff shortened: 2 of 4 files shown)


.github/workflows/trivy.yml
```diff
@@ -35,6 +35,7 @@ jobs:
 
       # build the image
       - name: Build
+        continue-on-error: true
```
jdcanas, line 38: Hmm.. what does this accomplish?

LizBaldo: This is just me testing the GHA, please ignore I will remove it soon

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -60,6 +61,23 @@ object DbReader {
           )"""
       .query[Runtime]
 
+  val deletingRuntimeQuery =
```
jdcanas, line 64: Seems like a different name is warranted since this gets runtimes in `Stopped` and `Running` as well. Maybe name it or leave a comment to indicate the usage as opposed to the statuses it includes.

LizBaldo: I am not sure I understand, the goal is to grab runtimes marked as `deleting` in leo db, but would show as Running or Stopped in Google. Similar to the deletedRuntimeQuery and the erroredRuntimeQuery

mlilynolting (approved):

.scalafmt.conf
```diff
@@ -1,4 +1,4 @@
-version = 3.6.1
+version = 3.7.12
```
Qi77Qi, line 1: thanks for keeping things up-to-date!

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
Qi77Qi: AoU re-uses runtime names...so just need to make sure we won't accidentally pick up runtimes that isn't in `Deleting`

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletingRuntimeChecker.scala
```diff
@@ -0,0 +1,104 @@
...
+import cats.effect.Concurrent
+import cats.mtl.Ask
+import cats.syntax.all._
+import org.broadinstitute.dsde.workbench.google2.DataprocClusterName
+import org.broadinstitute.dsde.workbench.model.TraceId
+import org.broadinstitute.dsde.workbench.util2.InstanceName
+import org.typelevel.log4cats.Logger
+
+// Implements CheckRunner[F[_], A]
+object DeletingRuntimeChecker {
```
Qi77Qi, line 13: maybe add a comment explaining what this checker intends to do

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -60,6 +61,23 @@ object DbReader {
...
+    sql"""SELECT DISTINCT c1.id, cloudContext, c1.cloudProvider, runtimeName, rt.cloudService, c1.status, rt.zone, rt.region
+          FROM CLUSTER AS c1
+          INNER JOIN RUNTIME_CONFIG AS rt ON c1.runtimeConfigId = rt.id
+          WHERE
+            c1.status = "Deleting" AND
+            c1.destroyedDate > now() - INTERVAL 60 MIN AND
```
Qi77Qi, line 70: should this be `c1.dataAccessed < now() - INTERVAL 60 MIN`...basically making sure the runtime hasn't been accessed for more than 1 hr.

`destroyedDate` will always be the dummy tmestamp becuz these runtimes aren't deleted yet

janitor/cloudbuild.yaml
```diff
@@ -2,7 +2,7 @@ steps:
   - name: 'gcr.io/$PROJECT_ID/scala-sbt'
     args:
       - |
-        testOnly -- -l cronJobs.dbTest; janitor/docker:publishLocal
+        JVM_OPTS="-Xmx4G -Xms4G -Xss2M"; testOnly -- -l cronJobs.dbTest; janitor/docker:publishLocal
```
LizBaldo, line 5: @Qi77Qi would you happen to know how I could pass these arguments to the docker container?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
LizBaldo: Good point, I will do that, thanks!

janitor/cloudbuild.yaml
Qi77Qi: try sth like 
```
  env:
  - JAVA_OPTS: -Xmx4G -Xms4G -Xss2M
 ```

see [this](https://cloud.google.com/build/docs/configuring-builds/substitute-variable-values) for setting env in cloud build

Comment (LizBaldo) [reply janitor/cloudbuild.yaml]: Ah thanks, looks slightly different from what I tried [here](https://github.com/broadinstitute/leonardo-cron-jobs/pull/289/commits/35e5e6db3c00d391280dfbad1a0f4bc3977f0eec), I will give it another try now :)

Comment (LizBaldo) [reply janitor/cloudbuild.yaml]: That worked, thanks a lot!

Comment (Qi77Qi) [reply janitor/cloudbuild.yaml]: yea, it should be `JAVA_OPTS`

**Comment by LizBaldo** [comment]
Fixed previous DBReader unit tests that are not run by CI but have to be manually done locally, and added a test for this specific use case, things work well with my Local leo DB now :)

File: core/src/main/scala/com/broadinstitute/dsp/checkerDeps.scala [hunk 7]
```diff
@@ -58,13 +63,14 @@ sealed abstract class Runtime {
 
 object Runtime {
   final case class AzureVM(id: Long,
-                           cloudContext: CloudContext.Azure,
+                           azureCloudContext: AzureCloudContext, //                           cloudContext: CloudContext.Azure,
```
rtitle, line 66: Remove comment?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -60,6 +61,23 @@ object DbReader {
...
       .query[Runtime]
 
+  val deletingRuntimeQuery =
+    sql"""SELECT DISTINCT c1.id, cloudContext, c1.cloudProvider, runtimeName, rt.cloudService, c1.status, rt.zone, rt.region
+          FROM CLUSTER AS c1
+          INNER JOIN RUNTIME_CONFIG AS rt ON c1.runtimeConfigId = rt.id
+          WHERE
+            c1.status = "Deleting" AND
+            c1.dateAccessed < now() - INTERVAL 1 HOUR AND
+            NOT EXISTS (
```
rtitle, line 71: Trying to understand why we need the `AND NOT EXISTS …` clause here. It seems unnecessary, since there is a unique constraint on `(cloudContext, name, destroyedDate)`?

LizBaldo: From Qi's comment above, it is necessary precaution as AOU reuses names

core/src/main/scala/com/broadinstitute/dsp/checkerDeps.scala
LizBaldo: Thanks!

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DeletingRuntimeChecker.scala
```diff
@@ -0,0 +1,104 @@
...
+        runtime match {
+          case x: Runtime.Dataproc =>
+            checkDataprocCluster(x, isDryRun)
+          case x: Runtime.Gce =>
+            checkGceRuntime(x, isDryRun)
+          case x: Runtime.AzureVM =>
+            checkAzureRuntime(x, isDryRun)
+        }
+
+      private def checkDataprocCluster(runtime: Runtime.Dataproc, isDryRun: Boolean)(implicit
```
rtitle, line 39: These methods seem the same as `DeletedRuntimeChecker`.. wonder if it could be made more DRY?

LizBaldo: I thought about it too, but the decoupling made is so easy to  develop and test that I ended up liking the framework and kept it, but if you feel strongly about it, I am happy to refactor

rtitle (commented):
This isn't necessarily new your PR, but something in the `resource-validator` SQL looks off to me.

1. `deletedRuntimeQuery`, `erroredRuntimeQuery`, `deletingRuntimeQuery` -- the action is the same in all these cases (we delete the runtime), so should they be collapsed into a single query + stream?

2. The `AND NOT EXISTS` is probably only necessary for the `Deleted` case -- otherwise the DB constraint prevents reusing names. And even the `deletedRuntimeQuery` SQL doesn't seem quite correct -- it only checks duplicates in Stopped and Running status, what if there is a runtime with the same name in `Starting`/`Stopping` while the RV happens top run?

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
rtitle: I think that would only be necessary for `Deleted` runtimes though, otherwise there is a DB constraint.

Qi77Qi: since the runtime is `Deleting`...I don't think `destroyedDate` will be defined yet... and (cloudContext, name) is not guranteed to be unique

rtitle: Yeah we only set `destroyedDate` when the status is `Deleted`… which means the constraint on `(cloudContext, name, destroyedDate)` only allows reusing names when at most 1 is non-deleted. (So you can't have a Deleting and Running runtime with the same name at the same time.)

I'm fine with leaving it as a sanity check… I was also wondering why the status check is not complete though. For example what if there were 2 runtimes:
```
name | status
-----
myRuntimeName | Deleted
myRuntimeName | Stopping
```

Would the `deletedRuntimeSql` mistakenly flag `myRuntimeName` and delete it in GCP?

Qi77Qi: I think the `NOT EXISTS` exclude `Stopped` as well (both `Stopped` and `Running` are "active" statuses basically)..

but yea, it's tricky...think we did this only becuz AOU reuses names...I'll bring it up to see if we can use unique names going forward for AOU

Qi77Qi: created https://precisionmedicineinitiative.atlassian.net/browse/RW-10776

Qi77Qi:
I'm fine with as is...hopefull we'll be able to clean this up a bit down the line

rtitle (approved):
I think my comments were a little beyond the scope of this PR, so thumbing this… would be great to clean up though.

PR #303, opened 2023-10-05
RW-10808 update APP_USAGE table

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-10808

When GKE cluster or nodepool gets deleted, we should make sure APP_USAGE table is updated to record `stopTime` if there's a record for the apps on these deleted cluster or nodepool. This PR updates `stopTime` to app's `dateAccessed` time, which is the last time Leonardo knows that a user accessed the app.

I tested the query against dev DB, and I think unit test has pretty good coverage.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,4 +32,4 @@ lazy val nuker = (project in file("nuker"))
   .enablePlugins(JavaAppPackaging)
   .dependsOn(core % "test->test;compile->compile")
 
-Test / parallelExecution := false
+concurrentRestrictions in Global += Tags.limit(Tags.Test, 1)
--- a/core/src/test/scala/com/broadinstitute/dsp/DbTestHelper.scala
+++ b/core/src/test/scala/com/broadinstitute/dsp/DbTestHelper.scala
@@ -29,7 +29,7 @@ object DbTestHelper {
   ): Resource[IO, HikariTransactor[IO]] =
     for {
       xa <- DbTransactor.init[IO](databaseConfig)
-      _ <- Resource.eval(truncateTables(xa))
+      _ <- Resource.make(IO.unit)(_ => truncateTables(xa))
     } yield xa
 
   def insertDiskQuery(disk: Disk, status: String) =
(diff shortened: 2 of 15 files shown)


core/src/test/scala/com/broadinstitute/dsp/DbTestHelper.scala
```diff
@@ -29,7 +29,7 @@ object DbTestHelper {
   ): Resource[IO, HikariTransactor[IO]] =
     for {
       xa <- DbTransactor.init[IO](databaseConfig)
-      _ <- Resource.eval(truncateTables(xa))
+      _ <- Resource.make(IO.unit)(_ => truncateTables(xa))
```
Qi77Qi, line 32: This is a small unrelated fix around unit tests.

Today, we can't run lots of the DB related unit tests because when they're run in bulk, they will fail for some reason. And I think this bug is one of the reasons (my own fault)...

So with this change, I'm able to remove `DbTest` tag all unit tests in `zombieMonitor`...I'll see if I can remove more if I have time

project/Dependencies.scala
```diff
@@ -6,7 +6,7 @@ object Dependencies {
   val workbenchGoogle2Version = s"0.32-${workbenchLibsHash}"
   val workbenchAzureVersion = s"0.5-${workbenchLibsHash}"
   val openTelemetryVersion = s"0.6-${workbenchLibsHash}"
-  val doobieVersion = "1.0.0-RC2"
+  val doobieVersion = "1.0.0-RC4"
```
LizBaldo, line 9: Is this why you shifted away from the HikariTransactor above?

LizBaldo (approved):
LGTM! I just have one small question, thanks for taking care of the dependencies update too :)

project/Dependencies.scala
Qi77Qi: kind of..we're still using `HikariTransactor` in the actual cron job code. It's just I'm making the code to use `Transactor` in test helpers so that in tests, we can use a different one for simplicity

build.sbt
```diff
@@ -32,4 +32,4 @@ lazy val nuker = (project in file("nuker"))
   .enablePlugins(JavaAppPackaging)
   .dependsOn(core % "test->test;compile->compile")
 
-Test / parallelExecution := false
+concurrentRestrictions in Global += Tags.limit(Tags.Test, 1)
```
Qi77Qi, line 35: @LizBaldo added this..this slows down tests run a bit...but I think it's ok since this repo is fairly small. The benefit is `sbt resourceValidator/test` will run successfully now :)

PR #319, opened 2024-01-31
[IA-4805] bug fix in zombie monitor/resource validator caused from last PR

Qi77Qi:
Tested resource validator and zombie monitor locally

### DIFF
--- a/core/src/main/scala/com/broadinstitute/dsp/DbReaderImplicits.scala
+++ b/core/src/main/scala/com/broadinstitute/dsp/DbReaderImplicits.scala
@@ -113,9 +113,18 @@ object DbReaderImplicits {
       case (id, name, cloudContextDb, location, cloudProvider) =>
         cloudProvider match {
           case CloudProvider.Azure =>
-            throw new RuntimeException(
-              s"kubernetesCluster(${id}) is Azure cluster. We should filter out Azure clusters in the query"
-            )
+            AzureCloudContext.fromString(cloudContextDb) match {
+              case Left(value) =>
+                throw new RuntimeException(
+                  s"${value} is not valid azure cloud context"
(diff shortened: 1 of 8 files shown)


core/src/main/scala/com/broadinstitute/dsp/DbReaderImplicits.scala
```diff
@@ -113,9 +113,18 @@ object DbReaderImplicits {
...
-            )
+            AzureCloudContext.fromString(cloudContextDb) match {
+              case Left(value) =>
+                throw new RuntimeException(
+                  s"${value} is not valid azure cloud context"
+                )
+              case Right(value) =>
+                KubernetesCluster(id, name, CloudContext.Azure(value), location)
+            }
+//           TODO: throw exception once we fix query in zombie monitor and resource validator
```
mlilynolting, line 124: Resource Validator does not support Azure yet, but Zombie Monitor does currently handle AKS clusters (see https://github.com/broadinstitute/leonardo-cron-jobs/blob/2fb1c0f194555c36269ed0888ff26762c1abcd14/zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedKubernetesClusterChecker.scala#L40). So I think that the exception is probably inappropriate here.

mlilynolting (commented):
This looks like a good fix. I'm unsure why the diff shows changes in DbReader; weren't those already done in your original PR?

mlilynolting:
Jira ticket is https://broadworkbench.atlassian.net/browse/IA-4805

mlilynolting (approved):

Qi77Qi:
@mlilynolting thanks for taking this to the finish line!

PR #321, opened 2024-03-06
[IA-4578] bump workbench-libs google2 version

mlilynolting:
Bump to include google2 library change https://github.com/broadinstitute/workbench-libs/pull/1669


LizBaldo:
Arg sorry I should have caught it as part of the workbenchlibs PR (I always forget myself, but you need to publish a new client for all the modules every time)

LizBaldo (approved):

PR #322, opened 2024-03-15
IA-4823  add destroyedDate condition

Qi77Qi:
We recently noticed that some PDs may have been marked as "Deleted" but without `destroyedDate` due to a bug. For these disks, cron job currently isn't validating whether they're indeed deleted in GCP, which can cause unexpected expense.

This PR adds these PDs to be scanned by the resource validator

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
@@ -32,7 +32,7 @@ object DbReader {
            (
              (pd1.status="Deleted" AND pd1.destroyedDate > now() - INTERVAL 30 DAY) OR
              (pd1.status="Deleting" AND pd1.`dateAccessed` < now() - INTERVAL 30 MINUTE) OR
-             (pd1.destroyedDate == ${leonardoDummyDate})
+             (pd1.status="Deleted" AND pd1.destroyedDate == ${leonardoDummyDate})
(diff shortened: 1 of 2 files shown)


resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -30,7 +31,8 @@ object DbReader {
            WHERE
            (
              (pd1.status="Deleted" AND pd1.destroyedDate > now() - INTERVAL 30 DAY) OR
-             (pd1.status="Deleting" AND pd1.`dateAccessed` < now() - INTERVAL 30 MINUTE)
+             (pd1.status="Deleting" AND pd1.`dateAccessed` < now() - INTERVAL 30 MINUTE) OR
+             (pd1.destroyedDate == ${leonardoDummyDate})
```
yonghaoy, line 35: Would that include other PD status like RUNNING?

yonghaoy (commented):
Also, do we need to hold off this change until we understand the impact: we need to know how long does they exist in GCP and how much credits we need to give back to users

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
Qi77Qi: oops...fixing

lucymcnatt (approved):

PR #330, opened 2024-03-18
IA-4871  Include destroyedDate in the deletedRuntime checker and fix a bug where we're including runtimes that shouldn't have been deleted

Qi77Qi:
1. Add back the `destroyedDate` condition from previous reverted PR
2. Fix a bug where we're detecting some runtimes as anomalies incorrectly.

Here's the bug. Take the following as an example
```
123099, all-of-us-6490, terra-vpc-sc-60fd067e, Deleted, 2023-11-12 00:48:55.071059, 2023-11-12 01:21:27.304278...
149253, all-of-us-6490, terra-vpc-sc-60fd067e, Stopping, 2024-03-18 00:48:55.071059, 1970-01-01 00:00:01.000000...
```

123099 should've been excluded becuz 149253 is in the same project with the same name that is still active (this is because AoU re-uses runtime names when creating new runtimes in the same project for the same user). But the query condition is not excluding 123099 because 149253's status is `Stopping` and the `NOT EXISTS` is only looking for `Stopped` and `Running`
```
 NOT EXISTS (
              SELECT *
              FROM CLUSTER AS c2
              WHERE
                c2.cloudContext = c1.cloudContext AND
                c2.runtimeName = c1.runtimeName AND
                (c2.status = "Stopped" OR c2.status = "Running"))
```

### DIFF
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
@@ -50,15 +50,18 @@ object DbReader {
           FROM CLUSTER AS c1
           INNER JOIN RUNTIME_CONFIG AS rt ON c1.runtimeConfigId = rt.id
           WHERE
-            c1.status = "Deleted" AND
-            c1.destroyedDate > now() - INTERVAL 30 DAY AND
+            (
+              (c1.status = "Deleted" AND c1.destroyedDate > now() - INTERVAL 30 DAY) OR
+              (c1.status="Deleted" AND c1.destroyedDate = '1970-01-01 00:00:01.000000')
+            )
+            AND
             NOT EXISTS (
               SELECT *
               FROM CLUSTER AS c2
               WHERE
                 c2.cloudContext = c1.cloudContext AND
                 c2.runtimeName = c1.runtimeName AND
-                (c2.status = "Stopped" OR c2.status = "Running")
+                (c2.status != "Deleted" OR c2.status != "Deleting")
           )"""
       .query[Runtime]
 
@@ -75,7 +78,7 @@ object DbReader {
               WHERE
                 c2.cloudContext = c1.cloudContext AND
                 c2.runtimeName = c1.runtimeName AND
-                (c2.status = "Stopped" OR c2.status = "Running")
+                (c2.status != "Deleted" OR c2.status != "Deleting")
           )"""
       .query[Runtime]
 
--- a/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
+++ b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
@@ -61,7 +61,8 @@ object DbReader {
               WHERE
                 c2.cloudContext = c1.cloudContext AND
                 c2.runtimeName = c1.runtimeName AND
-                (c2.status != "Deleted" OR c2.status != "Deleting")
+                c2.status != "Deleted" AND
+                c2.status != "Deleting"
           )"""
       .query[Runtime]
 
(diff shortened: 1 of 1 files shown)


yonghaoy (approved):

yonghaoy (approved):
LG. Only runtime would use the same name, hopefully we can get rid of this in the future then we can drop this

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
```diff
@@ -75,7 +79,8 @@ object DbReader {
               WHERE
                 c2.cloudContext = c1.cloudContext AND
                 c2.runtimeName = c1.runtimeName AND
-                (c2.status = "Stopped" OR c2.status = "Running")
+                c2.status != "Deleted" AND
+                c2.status != "Deleting"
```
cpate4, line 83: two questions:
- do we need to or want to modify any of the queries below?, and
- this looks like we are saying we want `status = deleting` matching the `dateAccessed` criteria above and excluding all non-`Deleted and Deleting` environments (which I think is another way of saying return all runtimes matching deleting and the date config) - which makes me want to ask, do we need this sub-query?

cpate4 (commented):
```scala
  CREATING("Creating"),
  RUNNING("Running"),
  UPDATING("Updating"),
  ERROR("Error"),
  STOPPING("Stopping"),
  STOPPED("Stopped"),
  STARTING("Starting"),
  DELETING("Deleting"),
  DELETED("Deleted"),  
  UNKNOWN("Unknown");
```
these are all of the ClusterStatus(es) from the generated leonardo-client (copying here so we know what we are dealing with)

questions asked as a part of the PR

resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DbReader.scala
Qi77Qi: >  do we need to or want to modify any of the queries below?
It might make sense to apply the same thing for `erroredRuntimeQuery` and `stoppedRuntimeQuery`...but I'd like to limit this PR to just what's necessary for now in case of any surprises.

> this looks like we are saying we want status = deleting matching the dateAccessed criteria above and excluding all non-Deleted and Deleting environments (which I think is another way of saying return all runtimes matching deleting and the date config) - which makes me want to ask, do we need this sub-query?
It's needed becuz AoU reuses runtime name for same project and user. See the example I listed in the description. Basically, there will be many "Deleted" or "Deleting" AoU runtimes with same project and name, and if there's an "active" runtime in the same project and name, then we want to skip these Deleted/Deleting runtimes, because if we don't skip, then we'll end up doing sth like `gcp.deleteRuntime(project, name)`, and it'll delete the active one while it shouldn't be deleted

cpate4: - 👍 on the first question, thank you. (makes sense), and
- makes sense re: name conflicts as you describe. do you know if because we are doing a `select *` in the `NOT EXISTS`-query and are including the `cluster.id in the global `SELECT ...`-statement above if we will actually be filtering out based on the cluster `name`? Asked another way, do we need to explicitly say `where name not in (insert-subquery-here)`?

Qi77Qi: >Asked another way, do we need to explicitly say where name not in (insert-subquery-here)`?
I don't think we need to...as long as `NOT EXISTS` returns anything, we'll exclude that runtime. So basically we're saying, for every `Deleting`/`Deleted` runtime, does there exist another one with the same name and same project, and its status is "active". If there is any such row in returned in the subquery, we'll exclude the runtime

cpate4: oh ... shoot ... my bad. I missed the `c1` amidst all the `c2`'s in the subquery. thanks for entertaining my questions!

cpate4 (approved):

PR #331, opened 2024-05-01
RW-12296 Added unit tests for zombieMonitor DeletedDiskChecker

evrii:
Before proceeding with changing Zombie Monitor behavior (as described [here](https://docs.google.com/document/d/1BR71eX5mWjXUoHRTYs0dMjGJe6eXa3K5sLO7neVn6So/edit)), I wanted to add some unit tests in order to verify current behavior. These will be helpful to test against when changing cron job behavior in a future ticket.

### DIFF
--- a/zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskCheckerSpec.scala
+++ b/zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskCheckerSpec.scala
@@ -70,7 +70,7 @@ class DeletedDiskCheckerSpec extends AnyFlatSpec with CronJobsTestSuite {
       res.unsafeRunSync()
 
       // Assert
-      verify(mockDbReader).updateDiskStatus(anyLong())
+      verify(mockDbReader).updateDiskStatus(disk.id)
     }
   }
   it should "not call updateDiskStatus with error 'billing is disabled'  (isDryRun = true)" in {
@@ -87,14 +87,14 @@ class DeletedDiskCheckerSpec extends AnyFlatSpec with CronJobsTestSuite {
         )
       )
 
-      when(mockDbReader.updateDiskStatus(anyLong())).thenReturn(IO.unit)
+      when(mockDbReader.updateDiskStatus(disk.id)).thenReturn(IO.unit)
 
       // Act
       val res = checker.checkResource(disk, isDryRun = true)
       res.unsafeRunSync()
 
       // Assert
-      verify(mockDbReader, never()).updateDiskStatus(anyLong())
+      verify(mockDbReader, never()).updateDiskStatus(disk.id)
     }
   }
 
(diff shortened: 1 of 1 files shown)


jdcanas:
Hmm... all the tests pass but there seems to be exception stacktraces in the test output that look fishy
https://github.com/broadinstitute/leonardo-cron-jobs/actions/runs/8942385738/job/24564865124?pr=331#step:3:142

evrii:
The exceptions you are seeing in the stack-trace are the exceptions that I am throwing as part of the test, so they should not be anything to be concerned with.

zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskCheckerSpec.scala
```diff
@@ -0,0 +1,261 @@
...
+    .build()
+  def setupMocks(): Unit = {
+    mockDbReader = mock(classOf[DbReader[IO]])
+    mockCheckRunnerDeps = mock(classOf[CheckRunnerDeps[IO]])
+    mockGoogleDiskService = mock(classOf[GoogleDiskService[IO]])
+    mockDiskCheckerDeps = DiskCheckerDeps(mockCheckRunnerDeps, mockGoogleDiskService)
+    checker = DeletedDiskChecker.impl(mockDbReader, mockDiskCheckerDeps)
+  }
+
+  it should "call updateDiskStatus with error 'billing is disabled' (isDryRun = false)" in {
```
lucymcnatt, line 52: the naming here was a little confusing to me at first, something like the following might be clearer?
`it should "updateDiskStatus when GCP returns a 'billing is disabled' error"`

zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskCheckerSpec.scala
```diff
@@ -0,0 +1,261 @@
...
+
+      // Act
+      val res = checker.checkResource(disk, isDryRun = false)
+      res.unsafeRunSync()
+
+      // Assert
+      verify(mockDbReader).updateDiskStatus(disk.id)
+    }
+  }
+  it should "not call updateDiskStatus with error 'billing is disabled'  (isDryRun = true)" in {
```
lucymcnatt, line 76: I'm not sure its necessary to have the `(isDryRun = true)` test for each of the error messages, if `isDryRun` is true it shouldn't update the diskStatus regardless of the error message. Rather than adding a test per exception type, it could be simpler to just have one `not updateDiskStatus if GCP returns an error in a dry run`

lucymcnatt: Since the behavior for the dry run is to never update the disk status, you could have one multi-test with different return types for the mock, something like the following might work:

```  
for (
    getDiskResponse <- List(
                IO.raiseError(billingDisabledException),
                IO.raiseError(computeEngineNotSetupException), 
                IO.pure(None)
)
    
it should s"never update disk status in a dry run" {
...
          .thenAnswer(_ => getDiskResponse)
...
}

```

evrii: I updated the code to consolidate the test code. I don't think that it is quite as readable, but it is certainly shorter.

jdcanas (approved):
Looks good pending Lucy's comments

evrii:
@NehaBroad FYI

zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskCheckerSpec.scala
lucymcnatt: looks good, thanks for making those changes!

lucymcnatt (approved):

PR #332, opened 2024-05-08
RW-12296 Changed cron job so that it no longer calls updateDiskStatus when the getDisk endpoint encounters a billing error.

evrii:
Follow up to https://github.com/broadinstitute/leonardo-cron-jobs/pull/331. 

Behavior of this cron job is that it currently updates disk status when it encounters a PermissionDenied exception with a message that contains either “Compute Engine API has not been used” or “This API method requires billing to be enabled” and isDryRun = false.

Both of these messages will no longer update disk status.

Tested this by running `sbt "zombieMonitor/run --checkDeletedDisks"`with cloud-sql container running, so the cron job ran against dev.

### DIFF
--- a/zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
+++ b/zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
@@ -36,12 +36,12 @@ object DeletedDiskChecker {
             case Left(e) =>
               val result = e match {
                 case ee: com.google.api.gax.rpc.PermissionDeniedException =>
-                  if (
-                    ee.getCause.getMessage.contains("Compute Engine API has not been used")
-                  ) {
+                  if (ee.getCause.getMessage.contains("Compute Engine API has not been used")) {
                     if (isDryRun) F.unit else dbReader.updateDiskStatus(disk.id)
-                  } else if(ee.getCause.getMessage
-                    .contains("This API method requires billing to be enabled")){
+                  } else if (
+                    ee.getCause.getMessage
+                      .contains("This API method requires billing to be enabled")
+                  ) {
                     logger.info(s"billing is disabled for ${disk.diskName}")
                   } else logger.error(e)(s"fail to get ${disk.diskName}")
                 case e => logger.error(e)(s"fail to get ${disk.diskName}")
--- a/zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
+++ b/zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
@@ -40,7 +40,7 @@ object DeletedDiskChecker {
                     ee.getCause.getMessage.contains("Compute Engine API has not been used") || ee.getCause.getMessage
                       .contains("This API method requires billing to be enabled")
                   )
-                    logger.info(s"failed to get ${disk.diskName} due to ${ee.getCause.getMessage}")
+                    logger.info(s"failed to get ${disk.diskName} due to ${ee.getCause.getMessage}") >> deps.checkRunnerDeps.metrics.incrementCounter("activeDiskBillingException")
                   else logger.error(e)(s"fail to get ${disk.diskName}")
                 case e => logger.error(e)(s"fail to get ${disk.diskName}")
               }
--- a/zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
+++ b/zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
@@ -36,14 +36,12 @@ object DeletedDiskChecker {
             case Left(e) =>
               val result = e match {
                 case ee: com.google.api.gax.rpc.PermissionDeniedException =>
-                  if (ee.getCause.getMessage.contains("Compute Engine API has not been used")) {
-                    if (isDryRun) F.unit else dbReader.updateDiskStatus(disk.id)
-                  } else if (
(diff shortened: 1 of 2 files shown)


zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
```diff
@@ -36,12 +36,14 @@ object DeletedDiskChecker {
...
-                      .contains("This API method requires billing to be enabled")
-                  )
+                  if (ee.getCause.getMessage.contains("Compute Engine API has not been used")) {
                     if (isDryRun) F.unit else dbReader.updateDiskStatus(disk.id)
-                  else logger.error(e)(s"fail to get ${disk.diskName}")
+                  } else if (
+                    ee.getCause.getMessage
+                      .contains("This API method requires billing to be enabled")
+                  ) {
+                    logger.info(s"billing is disabled for ${disk.diskName}")
```
evrii, line 45: If I remember correctly, when we (AoU and IA) met to talk about this change, I believe that we mentioned logging when billing is disabled might be useful instead of just removing the condition from the if statement. If anyone has different thoughts, please let me know.

zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskCheckerSpec.scala
```diff
@@ -72,49 +79,21 @@ class DeletedDiskCheckerSpec extends AnyFlatSpec with CronJobsTestSuite {
         val res = checker.checkResource(disk, isDryRun = false)
         res.unsafeRunSync()
 
-        // Assert
-        verify(mockDbReader).updateDiskStatus(disk.id)
-      }
-    }
-
-  for ((title, getDiskResponse) <- gcpFailureConditions)
```
evrii, line 80: These tests are no longer relevant, because isDryRun no longer effects what happens with these exceptions.

zombie-monitor/src/test/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskCheckerSpec.scala
```diff
@@ -170,4 +149,26 @@ class DeletedDiskCheckerSpec extends AnyFlatSpec with CronJobsTestSuite {
       verify(mockDbReader, never()).updateDiskStatus(disk.id)
     }
   }
+
+  it should "updateDiskStatus when a call to GCP's getDisk endpoint returns no disk (isDryRun = false)" in {
```
evrii, line 153: This test used to be tied to gcpFailureConditions, but since the other exceptions no longer trigger updateDiskStatus, this test seemed to make more sense on its own.

zombie-monitor/src/main/scala/com/broadinstitute/dsp/zombieMonitor/DeletedDiskChecker.scala
```diff
@@ -40,7 +40,7 @@ object DeletedDiskChecker {
                     ee.getCause.getMessage.contains("Compute Engine API has not been used") || ee.getCause.getMessage
                       .contains("This API method requires billing to be enabled")
                   )
-                    if (isDryRun) F.unit else dbReader.updateDiskStatus(disk.id)
+                    logger.info(s"failed to get ${disk.diskName} due to ${ee.getCause.getMessage}")
```
Qi77Qi, line 43: just occured to me, it'd be nice to add some metrics for this..just so we have a sense how things are going in the wild...

maybe we can chat abt it next mobbing or sth?

Qi77Qi: https://github.com/broadinstitute/leonardo-cron-jobs/blob/87b615eebf93143bf94c27466cbe9bd0687a888b/resource-validator/src/main/scala/com/broadinstitute/dsp/resourceValidator/DataprocWorkerChecker.scala#L72 is an example

evrii: That's a great idea. If you get a few minutes, I would love to talk with you about whatever setup is necessary to get that working.

LizBaldo, line 43: Maybe I am missing something, but from your PR description it sounded like you wanted to keep the updateDisk behavior when getting the `"Compute Engine API has not been used"` message? It looks like the disk won't be updated for that case either though

evrii: Sorry, I hadn't updated the description. After talking with @Qi77Qi . It sounds like updating shouldn't happen for either message.

This seems consistent with the conversation I found in an old PR relating to this:
https://github.com/broadinstitute/leonardo-cron-jobs/pull/320

LizBaldo: Gotcha, thanks for updating the descritption. In this case ping me when you have the logging piece and I can 👍 :)

evrii: The logging and metrics are all updated. Thanks for your review.

Qi77Qi: fwiw, I don't think cron jobs logs are in grafana..think they were previously in GCP, and we never ported them over to grafana.

Qi77Qi (approved):

LizBaldo (approved):

